PR #4, opened 2026-06-08
exchange_device_token trial and error

ChristopherHX:


xodus/src/api/live.rs
```diff
@@ -90,19 +276,49 @@ pub async fn exchange_device_token(
...
         .send()
         .await?;
 
     let text = response.text().await?;
+    println!("{}", text);
+
+        let result = bergshamra::verify(&ctx, &text).unwrap();
+        match result {
+            Invalid { reason } => {
+                print!("{}", reason);
```
ChristopherHX: @imLinguin I assume you mean this code, yes it is invalid and we need a new ctx with a correct key

imLinguin: gotcha

PR #5, opened 2026-06-10
Clep obfuscation cleanup

Vidi0:
Split the existing `clep_obfuscate` function into smaller pieces. Add a new `Cipher` struct in `clep::challenge` that stores the cipher state, and add associated functions to it that encrypt and update the internal state.

I tried to simplify the `clep_obfuscate` function because it did a lot of things in one single place. Now the function creates a new `Cipher` struct by obtaining the correct IV from the buffer and encrypts/obfuscates each block in place. Also, the `Cipher` struct now stores `Cipher::INITIAL_STATE` as a constant. It is being calculated with the same method as before, just in a separate const function.

I think overall the code is more readable now.

I tested a few examples of encrypting buffers with the old and new code and it produces the same exact outputs.

### DIFF
--- a/xodus/src/clep/challenge.rs
+++ b/xodus/src/clep/challenge.rs
@@ -1,6 +1,6 @@
 use crate::models::clep::*;
 
-use zerocopy::{FromZeros, transmute};
+use zerocopy::{FromZeros, transmute, transmute_mut};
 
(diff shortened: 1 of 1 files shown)


imLinguin (approved):
Thank you! This is much better now

PR #6, opened 2026-06-11
Apply Lint fixes

Vidi0:
Apply lint fixes as suggested by Cargo's Clippy.

The number of bytes read/written wasn't being taken into account, so fix that by using `read_exact` and `write_all`. Also, remove the `BufReader` in `SPLicense::from`, as `&[u8]` already implements `BufRead` directly.

In addition, fix the formatting (by running `cargo fmt`) and a few unused variable warnings.


Vidi0:
Rebased on top of main

imLinguin (approved):
Amazing, thanks

PR #10, opened 2026-06-11
feat: download via xsts tokens

ChristopherHX:
* request signature cannot be tested and is removed


xodus/src/api/xbox/mod.rs
```diff
@@ -0,0 +1,193 @@
...
+
+    let user_token = crate::api::live::exchange_user_token(
+        client,
+        legacy.token,
+        "USERNAME".to_string(),
+        dev_token.token,
+        secret,
+        None,
+        Some("Silent".to_string()),
+        "minecraft".to_string(),
```
imLinguin: Should we keep that hostingApp like this? hah

ChristopherHX: I think we need a common user token management later on, so that we always request all token variants at once.

Reverted this to the UUID we dumped and use for license

imLinguin (approved):

PR #11, opened 2026-06-11
Splicense cleanup

Vidi0:
Add `SPLicense::default()` and derive `TryFromPrimitive` for `BlockId`
Split the parse of a `SPLicense` into two functions: `decode`, which does the parsing itself; and `merge_tlv`, which merges a tag-length-value from the provided reader into the license. Add a few helper functions to simplify the tlv merging further.

Also, fix a small bug where the tag of each license element could be partly read. It's not really a bug in practice because a 4-byte buffer is almost always read fully, but in theory the reader could fill only part of it.

This PR only changes the behavior in one thing: before, a `BlockId::PackedContentKeys` `id_len` could be larger than 16 and the UUID would be read anyways, now the code asserts the `id_len` is exactly 16.

I hope these changes make the code a bit more readable and simple.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -5716,6 +5716,7 @@ dependencies = [
  "hmac 0.12.1",
  "keyring-core",
  "log",
+ "num_enum",
  "quick-xml",
  "rand 0.10.1",
  "reqwest",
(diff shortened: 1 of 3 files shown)


imLinguin (approved):
Good stuff!

PR #24, opened 2026-06-13
feat: add xodus-service ipc

imLinguin:
I want to kickoff the service work a little, I want to discuss the IPC aspect, and ideally start moving things to the new service crate

## List

- [x] init xodus-service tool
- [x] decide on connection flow with sdk
- [x] expose functionality for XUser
- [ ] expose functionality for XStore


imLinguin:
I think it makes sense to use protobuf for internal communication between processes. As I also aim to make CLI mostly just a service caller, that would enable adding the dedicated GUI app.

I imagine two sets of connections, one dedicated to the xgameservices.dll that is the best for communication with it and one for communication with native apps - protobuf based socket may be the best.

imLinguin:
Discussions from Discord: we decided to go with Unix socket

the packet would look like

```
magic: 4 bytes
message_size u16
message_type u16
message: String
```

message would be XML encoded

imLinguin:
Initial XUser funcs landed, currently for XML transport only.
I'll expand protobuf transport when its time to move CLI functionality to service

olivi-r:
From the XUser perspective, it is more useful to expose the User token rather then XSTS tokens, the public proof key used to sign messages is a member of `Properties` for XSTS requests.

Currently, to be able to sign requests, either XUser would need to send the public proof key to Xodus for every XSTS request, or Xodus would need to manage the keys and send the private signing key in the response. This key exchange could occur at a separate time to avoid the extra overhead, but there would still be a lot of overhead for every call to `XUserGetTokenAndSignature`, this is not a huge deal as the method is asynchronous anyways, but still not ideal.

In my opinion, both keys should be on the XUser side, if they were on the Xodus side, it would mean sending potentially quite large request data and context other ipc which would be a bottleneck.

imLinguin:
Lets review this initial iteration of the service. I'd like for it to evolve into a center of the whole Xodus utility. Ideally CLI and eventually GUI would connect to it, allowing for async dispatch of tasks. Right now the priority is to provide necessary context to xgameruntime to take over the authorization for given appid

ChristopherHX:
How does logging work? xodus-service --help didn't provide me any insight and was ignored.

imLinguin:
Right now CLI should be used to login. In the end I plan to have a flow where the service handles everything, CLI and other utils would just connect to it

ChristopherHX:
I meant log::debug writes into the void for me and have no clue to see it yet

imLinguin:
`XODUS_LOG=debug` should let you see those

ChristopherHX (approved):
Didn't come to actually test the service, but the xodus-cli works as before so this is good

ChristopherHX:
I confirm the python script works also on macOS, so ready for merge I guess?

PR #27, opened 2026-06-13
Add custom error enums in splicense

Vidi0:
Add custom error enums using the `thiserror` crate.

### DIFF
--- a/xodus-cli/src/commands/license.rs
+++ b/xodus-cli/src/commands/license.rs
@@ -1,7 +1,7 @@
 use crate::{device, user};
 use tokio::{fs::OpenOptions, io::AsyncWriteExt};
 use xodus::{
-    licensing::splicense::{parse_license, unpack_key},
+    licensing::splicense::{SPLicense, unpack_key},
     models::{live::ExchangeUserTokenOutcome, secrets::Token, soap},
 };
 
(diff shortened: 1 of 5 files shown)


xodus-cli/src/commands/license.rs
```diff
@@ -86,10 +86,10 @@ pub async fn run(client: &reqwest::Client, content_id: String, market: String, c
     .await
     .expect("failed to get license");
 
-    let game_splicense = parse_license(game_license.splicense_block);
+    let game_splicense = SPLicense::parse_base64(game_license.splicense_block).unwrap();
```
imLinguin: Could do expects instead to describe the fault better

ChristopherHX: I moved that code somewhere else, since extracts also needs the license...

Keep this in mind.

imLinguin: There is a conflict because of this now, indeed

Vidi0: I don't think using expect is a lot better, it still prints the debug error implementation instead of the display one.

imLinguin (approved):

ChristopherHX (approved):
later license.rs should also return these errors, it currently uses type String in my finite wisdom

PR #39, opened 2026-06-14
Forward to the login page when no token

### LINKED ISSUE #18: Download without valid session should forward to the login
If you have no active session and use the download command, you get an unhandled error.
Instead you should be warned and forwarded to the login screen

mandruis7:
Launch the login in case that there's no active sessions

Closes https://github.com/xodus-gaming/xodus/issues/18

### DIFF
--- a/xodus-cli/src/commands/download.rs
+++ b/xodus-cli/src/commands/download.rs
@@ -63,7 +63,7 @@ pub async fn run(client: &reqwest::Client, product: String, market: Option<Strin
         eprintln!("Invalid STS token");
         return;
     };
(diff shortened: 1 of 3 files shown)


imLinguin (changes requested):
I think we should handle the failure in specific cases, since we may want to use get_token to check if we even have the token locally, also keep in mind get_token can be used to various names we have stored, not just STS

imLinguin:
I'm going to close this for now. The login page error handling may be needed for refreshing stale tokens as well. So handling it will need more work

PR #41, opened 2026-06-16
Library Querying

Exterminate5573:
Adds a library command, currently only outputs raw json

### DIFF
--- a/xodus-cli/src/commands/library.rs
+++ b/xodus-cli/src/commands/library.rs
@@ -73,7 +73,7 @@ pub async fn run(client: &reqwest::Client) {
     }
     let xsts_header = api::xbox::get_xsts_auth_header(xsts.unwrap());
 
-    let games = api::xbox::services::get_library(client, user_token, xsts_header).await;
+    let games = api::xbox::services::get_library(client, user_token, xsts_header, None).await;
 
     println!("{:?}", games);
 }
--- a/xodus-cli/src/commands/library.rs
+++ b/xodus-cli/src/commands/library.rs
@@ -1,0 +1,79 @@
+use crate::{device, user};
+use xodus::models::secrets::Token;
+use xodus::api;
+use xodus::models::live::ExchangeUserTokenOutcome;
+use xodus::models::soap;
+
+pub async fn run(client: &reqwest::Client) {
+
+    let dev_token = device::get_device_token().unwrap();
+    let Token::Legacy(dev_token) = dev_token else {
+        eprintln!("Invalid STS token");
+        return;
+    };
+    let user = user::get_user().unwrap();
+    let user_token = user::get_token("http://Passport.NET/STS".to_string()).unwrap();
+    let Token::Legacy(legacy) = user_token else {
+        eprintln!("Unspported user token");
+        return;
+    };
+
+    let secret = dev_token.binary_secret.unwrap();
+
+    let user_token = api::live::exchange_user_token(
+        client,
+        legacy.token,
+        user.username,
+        dev_token.token,
+        secret,
+        None,
+        Some("Silent".to_string()),
+        "{d6d5a677-0872-4ab0-9442-bb792fce85c5}".to_string(),
+        &[(
+            "www.microsoft.com".to_owned(),
+            Some(soap::PolicyReference::mbi_ssl()),
+        )],
+    )
+        .await
+        .expect("Failed to get ms user token");
+
+    let user_token: Token = match user_token {
+        ExchangeUserTokenOutcome::Fault(_) => {
+            eprintln!("Failed to get exchange MS token");
+            return;
+        }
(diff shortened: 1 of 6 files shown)


xodus/src/api/xbox/services.rs
```diff
@@ -0,0 +1,115 @@
...
+    client: &reqwest::Client,
+    token: String,
+    xsts_header: String
+)  -> reqwest::Result<MyGames> {
+
+    let cv = CorrelationVector::new();
+
+    let resp = client
+        .get("https://beige.xboxservices.com/pcgafd/mygames")
+        .query(&[("market", "AU"), ("language", "en-US"), ("appVersion", "2606.1001.27.0")]) // TODO
```
imLinguin: market should be a param, or in the meantime hardcoded NEUTRAL

xodus-cli/src/commands/library.rs
```diff
@@ -0,0 +1,79 @@
...
+                                             token,
+                                         )) => token.into(),
+        _ => unreachable!("Only responses are handled"),
+    };
+    let Token::Compact(user_token) = user_token else {
+        eprintln!("Unspported token");
+        return;
+    };
+
+    let token = user::get_token("scope=service::user.auth.xboxlive.com::MBI_SSL&api-version=2.0".to_string()).unwrap();
```
imLinguin: The new token may need to be obtained here. Those tokens are pretty short-lived.

You can band-aid a full exchange_user_token here as well. Or if you want to, look into doing proper utils towards caching tokens and refreshing them only as needed

xodus-cli/src/commands/library.rs
```diff
@@ -0,0 +1,79 @@
...
+            "www.microsoft.com".to_owned(),
+            Some(soap::PolicyReference::mbi_ssl()),
+        )],
+    )
+        .await
+        .expect("Failed to get ms user token");
+
+    let user_token: Token = match user_token {
+        ExchangeUserTokenOutcome::Fault(_) => {
+            eprintln!("Failed to get exchange MS token");
```
imLinguin, line 41: For those eprintln, I think we should migrate to `log::error` instead

imLinguin (changes requested):

imLinguin:
I also hit

```
thread 'main' (130024) panicked at xodus/src/api/xbox/services.rs:116:57:
called `Result::unwrap()` on an `Err` value: Error("missing field `averageRating`", line: 1, column: 71440)
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace

```

xodus-cli/src/commands/library.rs
Exterminate5573: I dont think its worth band-aiding it all together, working on a proper caching system would make more sense

xodus/src/api/xbox/services.rs
```diff
@@ -0,0 +1,121 @@
...
+) -> reqwest::Result<MyGames> {
+    let cv = CorrelationVector::new();
+
+    let resp = client
+        .get("https://beige.xboxservices.com/pcgafd/mygames")
+        .query(&[
+            ("market", market.unwrap_or("neutral".to_string())),
+            ("language", "en-US".to_string()),
+            ("appVersion", "2606.1001.27.0".to_string()),
+        ]) // TODO
```
imLinguin, line 107: What is that TODO referring to?

xodus/src/api/xbox/services.rs
```diff
@@ -0,0 +1,121 @@
...
+    xsts_header: String,
+    market: Option<String>,
+) -> reqwest::Result<MyGames> {
+    let cv = CorrelationVector::new();
+
+    let resp = client
+        .get("https://beige.xboxservices.com/pcgafd/mygames")
+        .query(&[
+            ("market", market.unwrap_or("neutral".to_string())),
+            ("language", "en-US".to_string()),
```
imLinguin, line 105: Maybe worth for it to be param too

xodus-cli/src/commands/library.rs
```diff
@@ -0,0 +1,98 @@
...
+    };
+    if xsts.is_none() {
+        eprintln!("Failed to authenticate Xbox user");
+        return;
+    }
+    let xsts_header = api::xbox::get_xsts_auth_header(xsts.unwrap());
+
+    let games = api::xbox::services::get_library(client, user_token, xsts_header, None).await;
+
+    println!("{:?}", games);
```
imLinguin, line 97: Could the command display a nice list of titles and ProductIDs maybe?
log::debug!() could print the whole parsed payload

imLinguin (changes requested):

xodus/src/api/xbox/services.rs
Exterminate5573: The TODO should be 1 line up, it refers to the appVersion which shouldnt be a hardcoded string

PR #42, opened 2026-06-16
ci: add fmt and compilation checks

imLinguin:
no clippy yet, too early for that. We'll need to refactor some things nevertheless

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -1,0 +1,32 @@
+name: Rust check
+
+on:
+  pull_request:
+  push:
+    branches: [main]
+
+jobs:
+  fmt:
+    name: Check fmt
+    runs-on: ubuntu-slim
+    steps:
+      - uses: actions/checkout@v6
+      - run: cargo fmt --check --all
+  check:
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        include:
+          - target: x86_64-pc-windows-msvc
+            os: ubuntu-latest
+          - target: x86_64-unknown-linux-gnu
+            os: windows-latest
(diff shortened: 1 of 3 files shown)


.github/workflows/rust.yml
```diff
@@ -0,0 +1,32 @@
...
+    runs-on: ${{ matrix.os }}
+    strategy:
+      matrix:
+        include:
+          - target: x86_64-unknown-linux-gnu
+            os: ubuntu-latest
+          - target: aarch64-apple-darwin
+            os: macos-latest
+    steps:
+      - uses: actions/checkout@v6
```
Vidi0, line 25:
```suggestion
      - uses: actions/checkout@v6
      - uses: Swatinem/rust-cache@2
```

Vidi0: Adding a cache significantly decreases CI time. I suggest adding `Swatinem/rust-cache` because it is preconfigured to work best with Rust projects, but the default `actions/cache` could be used instead.

The format job doesn't need a cache.

.github/workflows/rust.yml
```diff
@@ -0,0 +1,32 @@
...
+      - name: Install dependencies
+        if: ${{runner.os == 'Linux'}}
+        run: sudo apt install libwebkit2gtk-4.1-dev
+      - name: Install toolchain
+        run: rustup target install ${{ matrix.target }}
+      - name: Run check
+        run: cargo check --target ${{ matrix.target }} --workspace
```
Vidi0: Adding the `--target` option is not needed if it is the current platform. Same for the `rustup target install`

Vidi0 (approved):

PR #44, opened 2026-06-17
Parse XVD structs

Vidi0:
Currently, XVD structs are being used directly using transmute from the raw bytes. This is very fragile because the structs aren't verified in any way. Also, it is very hard to work with those structs because the representation of their fields is raw, without parsing. For example, UUIDs are stored as [u8; 0x10] wihtout taking in account that they are mixed-endian GUIDs; and enums are stored as raw integers without checking if that integer makes sense.

This PR adds new structs which are smaller (by removing unknown/padding fields), are properly aligned (instead of the old packed ones), and have an endianness conversion for each integer larger than one byte (this wasn't a problem in practice because Xodus currently only runs on little-endian machines).

Furthermore, flags fields now use the `bitflags` crate instead of enums.

The conversion between arrays of bytes and the new XVD structs is done via macros because the Rust compiler can't currectly handle constant generics in that case.


xodus/src/xvd/utils.rs
```diff
@@ -225,177 +234,164 @@ pub struct EncryptedSectionInfo {
...
+                key_id,
+                offset,
+                length,
+                offset + length
+            );
 
-        if h.key_id != 0 {
-            continue;
-        }
+            if key_id.get().is_none() {
```
ChristopherHX: I bended the rules here, keyid = 0 is the first key, keyid 0xffff is none (do not quote me on count the f's). This code only supports a single key, all plain regions are not of interest here.

xodus/src/models/xvd/structs/parsed.rs
```diff
@@ -0,0 +1,508 @@
...
+        // `raw::XvcInfo` can hold up to 0xC0 encryption keys
+        // Any key higher than that means the region is unencrypted
+        if key_id < 0xC0 {
+            XvcKeyId(key_id as u8)
+        } else {
+            Self::UNENCRYPTED
+        }
+    }
+
+    fn is_unenctrypted(self) -> bool {
```
ChristopherHX: typo

xodus/src/models/xvd/structs/parsed.rs
```diff
@@ -0,0 +1,508 @@
...
+            value: value.value.map(|n| n.get()),
+        }
+    }
+}
+
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub struct XvcKeyId(u8);
+
+impl XvcKeyId {
+    const UNENCRYPTED: Self = Self(u8::MAX);
```
ChristopherHX: I remember that was more like 0xffff then

xodus/src/xvd/utils.rs
ChristopherHX: nvm, resolve this is fine

xodus/src/xvd/utils.rs
```diff
@@ -225,177 +234,164 @@ pub struct EncryptedSectionInfo {
...
-                false,
-                false,
-            );
-            let read_offset =
-                hash_tree_offset + page_number_to_offset(hash_block) + (entry_num * 0x18) + 0x14;
-            sfile.read_exact_at(&mut buf, read_offset).unwrap();
-            let u = u32::from_le_bytes(buf);
-            data_units.push(u);
-        }
+            let mut data_units: Vec<u32> = vec![];
```
ChristopherHX, line 349: we still need an assert than we do not decrypt key = 1 with key = 0, because this get awful behavior or skip them

xodus/src/models/xvd/structs/parsed.rs
Vidi0: Yes, the null key is 0xffff. However, I decided to count any key higher or equal to than 0xC0 (the number of key slots) as unencrypted. Then, to make the struct 1 byte instead of 2 the key is stored in a u8 instead of an Option<u8>. I had to choose one of the invalid values (from 0xC0 upwards) for the unencrypted value, so I took the max one. Therefore, this const is unrelated to how the XVD format stores invalid/unencrypted keys.

Vidi0: Maybe I should make this more clear

Vidi0: I decided to remove the optimization and store the key as an Option

xodus/src/xvd/utils.rs
Vidi0: It might make sense to support XVDs encrypted with multiple keys, but until then now it panics with a todo if it founds a key other than 0 or unencrypted.

Vidi0:
v2: Addressed comments.
Also, check magic bytes when converting into parsed struct representation.

xodus/src/xvd/utils.rs
```diff
@@ -225,177 +234,166 @@ pub struct EncryptedSectionInfo {
...
+                    xvd_header.number_of_hashed_pages(),
+                    start_page + page,
+                    0,
+                    false,
+                    false,
+                );
+                let read_offset = hash_tree_offset
+                    + page_number_to_offset(hash_block)
+                    + (entry_num * 0x18)
+                    + 0x14;
```
ChristopherHX: I think maybe we should create a struct of the hashtable entry being read here.

0x18 is the size, then we can reduce magic numbers in the future

0x14 => truncated sha256
0x4  => data that flows into decryption (seems to be an offsetted counter)

xodus/src/xvd/utils.rs
```diff
@@ -225,177 +234,166 @@ pub struct EncryptedSectionInfo {
...
+            0
+        } + hash_tree_offset;
+        let xvc_info_offset =
+            page_number_to_offset(xvd_header.user_data_page_count()) + user_data_offset;
+        let dynamic_header_offset =
+            page_number_to_offset(xvd_header.xvc_data_page_count()) + xvc_info_offset;
+        let drive_data_offset =
+            page_number_to_offset(xvd_header.dynamic_header_page_count()) + dynamic_header_offset;
+        let _dynamic_base_offset = xvc_info_offset;
+        let _static_data_length = match xvd_header.xvd_type {
```
ChristopherHX: TODO remove the unused variables, they are a leftover from my side.

ChristopherHX (approved):
I think this is good 👍 , rebase my branch then sometime later.

Looking at utils.rs with hidden whitespace changes looks simple.

xodus/src/xvd/utils.rs
Vidi0: I added the struct as requested 👍. It is now `XvdHashEntry`

Vidi0:
@ChristopherHX
I added the `XvdHashEntry` struct. I don't know if it is something like you had in mind, but it seems to keep working as before.

ChristopherHX:
Yes this was what I had in mind, thanks.

<details><summary>In my other PR I also had these new structs, but you do not need add them here</summary>

```rust
#[derive(FromBytes)]
#[repr(C, packed)]
pub struct XvdUserDataHeader {
    pub length: u32,
    pub version: u32,
    pub t: u32,
    pub unknown: u32,
}

#[derive(FromBytes)]
#[repr(C, packed)]
pub struct XvdUserDataPackageFilesHeader {
    pub version: u32,
    pub package_full_name: [u16; 260], // UTF-16
    pub file_count: u32,
}

#[derive(FromBytes)]
#[repr(C, packed)]
pub struct XvdUserDataPackageFileEntry {
    pub file_path: [u16; 260], // UTF-16
    pub size: u32,
    pub offset: u32,
}

#[derive(FromBytes)]
#[repr(C, packed)]
pub struct XvdSegmentMetadataHeader {
    pub magic: u32,
    pub version0: u32,
    pub version1: u32,
    pub header_length: u32,
    pub segment_count: u32,
    pub file_paths_length: u32,
    pub pduid: [u8; 0x10],
    pub unknown: [u8; 0x3c],
}

#[derive(FromBytes)]
#[repr(C, packed)]
pub struct XvdSegmentMetadataSegment {
    pub flags: u16,
    pub path_length: u16,
    pub path_offset: u32,
    pub filesize: u64,
}
```

</details>

They store an optional file list of an data segment

imLinguin (approved):
I really like these changes

PR #46, opened 2026-06-18
XVD: simplify `decrypt_page_xts` and `gf_mul_x`

Vidi0:
Optimize the `gf_mul_x` function, which is now 2x faster. Move it to the `math` module.

<details>
<summary>Micro-benchmark:</summary>
The test benchmark ran the function 10M times.

```
hyperfine "old" "new"
Benchmark 1: old
  Time (mean ± σ):      17.7 ms ±   1.2 ms    [User: 17.3 ms, System: 0.5 ms]
  Range (min ... max):    16.7 ms ...  19.9 ms    146 runs

Benchmark 2: new
  Time (mean ± σ):       8.8 ms ±   0.3 ms    [User: 8.4 ms, System: 0.5 ms]
  Range (min ... max):     8.2 ms ...   9.6 ms    308 runs

Summary
  new ran
    2.01 ± 0.15 times faster than old
```
</details>

I don't know if the performance benefit will be measurable at all in the decrypt function (it was already very fast, and the function wasn't called that many millions of times per second), but the function is now simpler. Also, add some tests and doc comments.

Simplify the `decrypt_page_xts` (and probably make it a bit faster, not benchmarked) and use integers to store the AES blocks instead of arrays to make the XORing simpler.


imLinguin:
This sounds like something that optimizations were doing already, release builds were substantially faster. Did you do any measurements on that release config?

Vidi0:
The benchmarks for `gl_mul_x` where it is 2x faster were done in release mode. I checked the assembly and now it is equally as fast in release and debug modes, before it was only fast in release mode.

However, I don't think this optimization really adds any noticeable performance to the extraction itself, the AES block decryption takes a lot more CPU cycles than this multiplication. Still, I think this PR is nice in that it cleans up the code and simplifies it.

ChristopherHX:
_I cannot say I fully understand the xvdtool logic this was based on yet_, for me is network infinite expensive and CPU very cheap right now_

I will have a look if the 20s extracting get's faster with this.

Vidi0:
I don't think it will get any faster. We could probably get a 6x-8x speedup if we parallelized the decryption of blocks in each CPU core (I would like to do it after a few refactors).

Vidi0:
<details>

<summary>New benchmarks</summary>

I am not sure about the veracity of these benchmarks, as the function signatures were quite different, so I tried my best to treat both functions equally.

This is benchmarking the `decrypt_page_xts` function by running it one million times. `bench2` is before this PR, and `bench1` is the current state of this PR.

```
hyperfine "./bench1" "./bench2" --warmup 5
Benchmark 1: ./bench1
  Time (mean ± σ):     459.1 ms ±   0.9 ms    [User: 457.2 ms, System: 0.7 ms]
  Range (min … max):   457.6 ms … 460.7 ms    10 runs
 
Benchmark 2: ./bench2
  Time (mean ± σ):      2.087 s ±  0.004 s    [User: 2.081 s, System: 0.001 s]
  Range (min … max):    2.082 s …  2.095 s    10 runs
 
Summary
  ./bench1 ran
    4.55 ± 0.01 times faster than ./bench2
```

</details>

Vidi0:
More benchmarks: this PR improves extraction time in Debug mode more than 3x

imLinguin (approved):
Great improvements!

PR #47, opened 2026-06-20
fix: hashtable read performance

ChristopherHX:
Group hashtable reads into larger reads via memory cache.

---

Background:

I faced huge delays in remote reading the hashtable, disk backed xvd this still saves 6s for me. On network this are multiple minutes for me.

before
       20,28 real         7,73 user        10,73 sys
after
       13,61 real         5,80 user         4,95 sys


xodus/src/xvd/utils.rs
```diff
@@ -349,24 +350,36 @@ impl XvdFile {
...
-                    + (entry_num * XvdHashEntry::RAW_SIZE as u64);
-
+                    + (entry_start * XvdHashEntry::RAW_SIZE as u64);
                 file.seek(SeekFrom::Start(read_offset)).await?;
-
-                let Ok(hash) = read_struct!(XvdHashEntry, file);
-                data_units.push(hash.unit);
+                let mut buf = vec![];
+                buf.resize(run_length as usize * XvdHashEntry::RAW_SIZE, 0);
+                file.read_exact(buf.as_mut_slice()).await.unwrap();
```
ChristopherHX: Further ideas, move out of loop.

Use const sized array with maximal possible hash run length e.g. 0xAA. And other ideas

xodus/src/xvd/math.rs
```diff
@@ -28,27 +28,28 @@ pub fn page_number_to_offset(page_number: u64) -> u64 {
...
     hash_tree_levels: u64,
     number_of_hashed_pages: u64,
     block_num: u64,
     hash_level: u32,
     resilient: bool,
     unknown: bool,
-) -> (u64, u64) {
+) -> (u64, u64, u64) {
     fn hash_block_exponent(block_count: u32) -> u64 {
         0xAAu64.pow(block_count)
```
ChristopherHX: move 0xAA to a public const in xvd model?

ChristopherHX: PAGES_PER_BLOCK we had that constant already

imLinguin:
are those reads close to one another? maybe a good idea would be to simply use a BufReader to avoid many syscalls. Instead of implementing that yourself with Cursor

ChristopherHX:
> are those reads close to one another

yes very very close. Didn't know BufReader exists, just searched for a reader over a vec!.

ChristopherHX:
Thanks for the hint, BufReader works and is exactly as fast. Updated the code.

ChristopherHX:
Actually this is not only iops, this function is heavy calculate_hash_block_num_for_block_num. My change calls it 0xaa times less than before.

Maybe eventually look into advance the hashblock location by other calculation if it is still a performance blocker.

17,91 real         6,72 user         8,18 sys, is the saving without bufreader.

xodus/src/xvd/math.rs
```diff
@@ -1,7 +1,10 @@
-use crate::models::xvd::constants::{
-    BLOCK_SIZE, DATA_BLOCKS_IN_LEVEL0_HASHTREE, DATA_BLOCKS_IN_LEVEL1_HASHTREE,
-    DATA_BLOCKS_IN_LEVEL2_HASHTREE, DATA_BLOCKS_IN_LEVEL3_HASHTREE, HASH_ENTRIES_IN_PAGE,
-    LEGACY_SECTOR_SIZE, PAGE_SIZE, SECTOR_SIZE,
+use crate::models::xvd::{
+    PAGES_PER_BLOCK,
```
Vidi0: Import PAGES_PER_BLOCK alongside the rest of constants

xodus/src/xvd/utils.rs
```diff
@@ -349,24 +354,32 @@ impl XvdFile {
             let mut data_units: Vec<u32> = vec![];
```
Vidi0: This could probably be optimized further by preallocating the `data_units` vector with the correct capacity.

```suggestion
            let mut data_units: Vec<u32> = Vec::with_capacity(xvd_header.number_of_hashed_pages() as usize);
```

Edit: number_of_hashes_pages is too high, but maybe preallocating some capacity is useful.

Vidi0: Actually, move the `data_units` definition down a few lines and add:

```
let mut data_units: Vec<u32> = Vec::with_capacity(num_pages as usize);
```

ChristopherHX:
Need to address the review this week, was busy with my fork of that code.

We may want to remove data_units vector and store both hash and dataunit per entry instead of two vectors in my current draft <https://discord.com/channels/1123890623586504714/1123953698440220672/1518732229944283188>

ChristopherHX:
Applied both suggestions

imLinguin (approved):

Vidi0 (approved):

PR #49, opened 2026-06-24
XVD: `decrypt_page_xts` optimization

Vidi0:
Continues #46

Optimize `decrypt_page_xts` a bit further by creating an iterator over the tweaks, and then zipping both the block and the tweak into the for loop. This way, no mutable state from outside the loop is needed in the iteration, making the code more readable (in my opinion) and simpler for LLVM to optimize.

By computing the next tweak before the AES decryption, and due to some complicated CPU stuff that I don't deeply understand, the tweak is calculated at the same time as the AES decryption, making the function a bit faster (around 8%).

I tried other versions of the function, like decrypting multiple AES blocks in parallel in the same CPU core or precomputing all tweaks into an array, but no version was faster than the one in this PR.

Also, the `decrypt_page_xts` function now receives the page by mutable reference instead of receiving it by value because it was around 12% faster in benchmarks. I wasn't able to check, however, whether that optimization was already being done automatically by LLVM in Xodus.

Benchmark:
```
hyperfine --warmup 10 ./bench1-1 ./bench1-2
Benchmark 1: ./bench1-1
  Time (mean ± σ):     464.7 ms ±   1.3 ms    [User: 461.9 ms, System: 1.3 ms]
  Range (min … max):   463.3 ms … 467.3 ms    10 runs
 
Benchmark 2: ./bench1-2
  Time (mean ± σ):     430.7 ms ±   3.8 ms    [User: 428.5 ms, System: 0.8 ms]
  Range (min … max):   426.7 ms … 439.2 ms    10 runs
 
Summary
  ./bench1-2 ran
    1.08 ± 0.01 times faster than ./bench1-1
```

### DIFF
--- a/xodus/src/xvd/crypt.rs
+++ b/xodus/src/xvd/crypt.rs
@@ -3,6 +3,7 @@ use crate::models::xvd::{PAGE_SIZE, XvcRegionId};
 use crate::xvd::math::gf_mul_x;
 
 use std::io::{self, Read, Seek, SeekFrom};
+use std::iter;
 
 use aes::Aes128;
 use aes::cipher::{BlockCipherDecrypt, BlockCipherEncrypt, KeyInit};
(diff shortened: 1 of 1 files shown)


ChristopherHX:
I need also an encrypt variant of that (one line off), or we cannot repair games on disk without full redownload.

so I had once a boolean to switch aes encrypt to decrypt

Vidi0:
Added an `encrypt_page_xts` function as requested 👍

ChristopherHX (approved):
works for me and looks good, rebase my thing on this so I need it in shortly

ChristopherHX:
btw 11s extracting with all files locally now for me using Minecraft GDK 26.21

ChristopherHX:
We need to coordinate this with https://github.com/xodus-gaming/xodus/pull/40, which makes the some crypt methods public. As streaming skips recreating aes on every 4k read.

Vidi0:
@ChristopherHX Now the decrypt and encrypt functions are public. If you want I could merge this now.

ChristopherHX:
let's merge

PR #51, opened 2026-06-27
feat: XSP format support

imLinguin:
Parsing of the file works great, some fields are still not quite mapped, but we are at the stage where we can reliably read the body of XSP file and know which blocks were modified or copied to different offset


msixvc/src/models/xsp/structs/parsed.rs
```diff
@@ -0,0 +1,97 @@
...
+    pub record_count: u32,
+    pub total_download: u64,
+    pub disk_space_required: u64,
+    pub upgrade_from_version: Version,
+    pub upgrade_to_version: Version,
+}
+
+#[derive(thiserror::Error, Debug)]
+pub enum XspHeaderParseError {
+    #[error(r#"invalid magic: expected "MS-XPFM": {0:?}"#)]
```
Vidi0:
```suggestion
    #[error(r#"invalid magic: expected "MS-XPFM ": {0:?}"#)]
```
Missing a space there

msixvc/src/common.rs
```diff
@@ -0,0 +1,61 @@
...
+    // Unix time counts nanoseconds since 1970-01-01 UTC.
+
+    /// Number of 100 nanoseconds between FILETIME epoch and Unix time
+    const FILETIME_TO_UNIX: i64 = 116_444_736_000_000_000;
+
+    let unix_nanos = (filetime - FILETIME_TO_UNIX) * 100;
+    DateTime::from_timestamp_nanos(unix_nanos)
+}
+
+pub trait Sealed {}
```
Vidi0:
```suggestion
```

The idea was that the `Sealed` trait couldn't be implemented from outside the module, avoiding implementing the `XvcStruct` trait from anything but the macro. However, if the macro has to be public to reuse the code for XVD and XSP, it might make more sense to just remove the sealed trait.

msixvc/src/common.rs
```diff
@@ -0,0 +1,61 @@
...
+
+/// An XVD binary structure that can be decoded from bytes.
+///
+/// The conversion from bytes isn't implemented as a trait method because the compiler
+/// hates generic constants: "generic parameters may not be used in const operations".
+///
+/// Implementing this trait means that `impl TryFrom<[u8; XvdStruct::RAW_SIZE]>
+/// for Self` exists.
+pub trait XvcStruct: Sized + Sealed {
```
Vidi0: I would move the XvcStruct trait and the macros to `msixvc::models::common`

msixvc/src/models/common.rs
Vidi0, line 1: I would prefer having this on msixvc::models::common (after all, it's just a portion of the parsed struct)

msixvc/src/utils.rs
Vidi0, line 1: I would remove the nesting and just move the `xsp` and `xvd` modules into the root of `msixvc`.

Vidi0 (approved):

msixvc/src/xsp.rs
```diff
@@ -0,0 +1,33 @@
...
+}
+impl XspFile {
+    pub async fn parse_file(path: String) -> Result<Self, Box<dyn std::error::Error>> {
+        let file = OpenOptions::new()
+            .read(true)
+            .open(path.clone())
+            .await
+            .expect("Unable to open file");
+
+        let mut file = BufReader::new(file);
```
ChristopherHX, line 15: please follow my pattern
```
    pub async fn parse_file(path: String) -> Result<Self, Box<dyn std::error::Error>> {
        let mut file = OpenOptions::new().read(true).open(path.clone()).await?;
        Self::parse(&mut file).await
    }

    pub async fn parse<Reader>(file: &mut Reader) -> Result<Self, Box<dyn std::error::Error>>
    where
        Reader: AsyncRead + AsyncSeek + Unpin,
    {
```

allow to provide an reader from any source

ChristopherHX (approved):
_better not delay too much, you refactor files I work with..._

msixvc/src/xsp.rs
imLinguin: good point

PR #52, opened 2026-06-27
Move all XVD related modules to a new `msixvc` crate

Vidi0:
Creates a new crate in the workspace, `msixvc`, and removes those files from the main Xodus lib crate. This way, it's more clear that what those two crates do is different, and that there is no relationship between the two crates. The only dependency that `msixvc` had in the main `xodus` crate was importing the `ContentKey` struct, but it can be passed as an array directly.

Also, this helps compile times, because the `msixvc` crate doesn't need to be rebuilt if only the `xodus` crate changes, and vice versa.

Fixed the lint, too.


xodus-cli/src/commands/extract.rs
```diff
@@ -28,6 +26,6 @@ pub async fn run(
     }
     if let Some((_, content_key)) = game_splicense.content_keys.into_iter().next() {
         let unpacked = content_key.unpack(&key).expect("failed to unpack");
-        unpack_file(xvd, path.to_string(), destination.to_string(), unpacked).expect("unpack ok");
+        unpack_file(xvd, path.to_string(), destination.to_string(), *unpacked).expect("unpack ok");
```
ChristopherHX, line 29: Ah with `*unpacked` you remove the type wrapper of ContentKey?

Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [workspace]
 resolver = "3"
-members = ["xodus", "xodus-cli", "xodus-service"]
+members = ["msixvc","xodus", "xodus-cli", "xodus-service"]
```
ChristopherHX:
```suggestion
members = ["msixvc", "xodus", "xodus-cli", "xodus-service"]
```

xodus-cli/src/commands/streaming.rs
```diff
@@ -64,14 +62,13 @@ pub async fn run(
...
-            eprintln!("{}", err.to_string());
+            eprintln!("{}", err);
             return;
         };
         let Some(file) = package
             .package_files
             .iter()
-            .filter(|p| p.file_name.ends_with(".msixvc"))
-            .next()
+            .find(|p| p.file_name.ends_with(".msixvc"))
```
ChristopherHX, line 71: didn't know that this works magically... 👀

ChristopherHX (approved):
Looks good so far...

xodus-cli/src/commands/extract.rs
Vidi0: Yes, this way the `msixvc` crate doesn't depend on `xodus`. The `*` dereferences the ContentKey struct into the raw array.

Cargo.toml
Vidi0: Good catch!

PR #56, opened 2026-06-29
chore: include basic ids on gamepass subs

imLinguin:
This will come in handy for entitlement detection

### DIFF
--- a/xodus/src/models/xbox.rs
+++ b/xodus/src/models/xbox.rs
@@ -1,3 +1,5 @@
+pub mod subscriptions;
+
 use serde::{Deserialize, Serialize};
 
 #[derive(Debug, Serialize)]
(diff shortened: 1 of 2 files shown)


ChristopherHX:
> entitlement detection

In what sense? I found such gamepass ids in the download package query we do in licensed by gamepass a, b, c, Minecraft Java, Minecraft for Windows etc.

_I am not entitled for gamepass, so for me they could be used to filter gamepass ids from the bundle explorer I barely added_

e.g. if you use prodid Minecraft Java and Bedrock you currently see noise of gamepass...

imLinguin:
I mean more in context of calling the endpoint - enumerating the IDs here, to know which subscription user owns. A UI could list all games they are entitled to with a few more queries

ChristopherHX (approved):

ChristopherHX:
I wonder if there is an api to list your subscription without using ids at all.

The catalog endpoints without auth are not trustworthy for this kind.

imLinguin:
I know of one API endpoint that can be used to check subscription status, the client sends those IDs however

PR #57, opened 2026-06-29
xvd: extract: speed up by avoiding data unit clone

Vidi0:
Extracting a MSIXVC file was slow because the large `data_units` vector was being cloned constantly. Replace the `Vec` clone with an `Rc` reference counted clone, which is much faster.

This is kind of a hack, it's not bad but in the end I would like to avoid cloning the data units at all.

Benchmarks:
```
Benchmark 1: ./xodus-old extract Microsoft.MinecraftUWP_1.26.3005.0_x64__8wekyb3d8bbwe.msixvc Microsoft.MinecraftUWP_1.26.3005.0_x64__8wekyb3d8bbwe/
  Time (mean ± σ):     10.054 s ±  0.147 s    [User: 6.097 s, System: 2.338 s]
  Range (min … max):    9.829 s … 10.283 s    10 runs
 
Benchmark 2: ./xodus-new extract Microsoft.MinecraftUWP_1.26.3005.0_x64__8wekyb3d8bbwe.msixvc Microsoft.MinecraftUWP_1.26.3005.0_x64__8wekyb3d8bbwe/
  Time (mean ± σ):      5.982 s ±  0.673 s    [User: 0.836 s, System: 1.939 s]
  Range (min … max):    4.925 s …  6.809 s    10 runs
 
Summary
  ./xodus-new extract Microsoft.MinecraftUWP_1.26.3005.0_x64__8wekyb3d8bbwe.msixvc Microsoft.MinecraftUWP_1.26.3005.0_x64__8wekyb3d8bbwe/ ran
    1.68 ± 0.19 times faster than ./xodus-old extract Microsoft.MinecraftUWP_1.26.3005.0_x64__8wekyb3d8bbwe.msixvc Microsoft.MinecraftUWP_1.26.3005.0_x64__8wekyb3d8bbwe/
```

### DIFF
--- a/msixvc/src/crypt.rs
+++ b/msixvc/src/crypt.rs
@@ -35,7 +35,7 @@ impl Tweak {
     }
 }
 
-pub struct SectionReader<R,'t> {
+pub struct SectionReader<'t, R> {
     inner: R,
     section_offset: u64,
     section_length: u64,
@@ -54,7 +54,7 @@ pub struct SectionReader<R,'t> {
     cached_page_plaintext: [u8; PAGE_SIZE],
 }
 
-impl<R: PageSource, 't> SectionReader<R> {
+impl<'t, R: PageSource> SectionReader<R> {
     pub fn new(
         inner: R,
         section_offset: u64,
--- a/msixvc/src/crypt.rs
+++ b/msixvc/src/crypt.rs
@@ -3,6 +3,7 @@ use crate::models::xvd::{PAGE_SIZE, XvcRegionId};
 
 use std::io::{self, Read, Seek, SeekFrom};
 use std::iter;
+use std::rc::Rc;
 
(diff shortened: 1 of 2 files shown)


ChristopherHX:
use non rc references in sectionreader? Since rc is not threadsafe, better let the compiler prevent anything needing complex lifetime

e.g. `<'t>` references? Like some newer code from me is doing.

I actually want to delete sectionreader, given it's cache is noop at the moment.

Vidi0:
Yeah, that's better. But let me change the reference to a Vec to a reference to a slice.

ChristopherHX:
yes feel free to make it again a slice, I approve now.

ChristopherHX (approved):

imLinguin (approved):

PR #60, opened 2026-07-01
cargo/config.toml: add aes target feature

Vidi0:
Add a note in the README that decryption of MSIXVC files can be much faster by compiling with the `aes` target feature. See the benchmarks below for more information.

We should compile with the `aes` target feature enabled when releasing compiled binaries. Without it, decryption is slow enough to block the tokio runtime, requiring `spawn_blocking` instead of running inline in an async task.

The main drawback is that the `aes` feature is not available on very old CPUs, but then the decryption would fall back to software, and that is a lot slower, I would say unusable (see the benchmarks).

Edit: enable the features by default by adding them to `.cargo/config.toml`
 
## Benchmarks
Tested on an `AMD Ryzen 5 7600`
```
hyperfine --warmup 2 ./bench-native ./bench-aes ./bench-none ./bench-soft
Benchmark 1: ./bench-native
  Time (mean ± σ):     383.6 ms ±   1.5 ms    [User: 381.7 ms, System: 0.9 ms]
  Range (min … max):   381.3 ms … 386.2 ms    10 runs
 
Benchmark 2: ./bench-aes
  Time (mean ± σ):     439.7 ms ±   4.0 ms    [User: 437.8 ms, System: 0.9 ms]
  Range (min … max):   435.5 ms … 444.2 ms    10 runs
 
Benchmark 3: ./bench-none
  Time (mean ± σ):      4.007 s ±  0.002 s    [User: 3.998 s, System: 0.001 s]
  Range (min … max):    4.004 s …  4.008 s    10 runs
 
Benchmark 4: ./bench-soft
  Time (mean ± σ):     64.137 s ±  0.130 s    [User: 63.988 s, System: 0.001 s]
  Range (min … max):   63.987 s … 64.372 s    10 runs
 
Summary
  ./bench-native ran
    1.15 ± 0.01 times faster than ./bench-aes
   10.45 ± 0.04 times faster than ./bench-none
  167.20 ± 0.72 times faster than ./bench-soft
```

| Benchmark | Compile flags | Time per block | Throughput |
|-----------|---------------|----------------|------------|
| `bench-native` | `-Ctarget-cpu=native` | ~383.6 ns | ~10.4 GiB/s |
| `bench-aes` | `-Ctarget-feature=+aes,+ssse3` | ~439.7 ns | ~9.1 GiB/s |
| `bench-none` | _(none)_ | ~4.0 µs | ~1.0 GiB/s |
| `bench-soft` | software fallback (reference) | ~64.1 µs | ~62.4 MiB/s |

### DIFF
--- a/.cargo/config.toml
+++ b/.cargo/config.toml
@@ -1,0 +1,8 @@
+# See https://docs.rs/aes/0.9.1/aes/#supported-backends
+# and https://github.com/xodus-gaming/xodus/pull/60
+
+[target.'cfg(target_arch = "x86_64")']
+rustflags = ["-Ctarget-feature=+aes,+ssse3"]
+
+[target.'cfg(target_arch = "aarch64")']
+rustflags = ["-Ctarget-feature=+aes"]
--- a/README.md
+++ b/README.md
@@ -86,6 +86,15 @@ Running xodus-service in debug
 cargo run --bin xodus-service
 ```
 
+> [!TIP]
+> For better performance when decrypting MSIXVC files, enable the `aes` target feature by passing
+> `RUSTFLAGS=-Ctarget-feature=+aes,+ssse3` to cargo build.
+>
+> See https://docs.rs/aes/0.9.1/aes/#x86x86_64-intrinsics-aes-ni-and-vaes for more information.
+>
+> Note that AES-NI is not available on very old CPUs.
+> See https://en.wikipedia.org/wiki/AES_instruction_set#x86_architecture_processors for a list of compatible CPUs.
+
 ### CLI Usage
 
 ```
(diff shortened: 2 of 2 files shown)


imLinguin:
Should we consider doing this by default? Most CPUs nowadays should support AES instructions

ChristopherHX:
> -Ctarget-feature=+aes,+ssse3

Sounds very x86_64 cpu specific, but not explicitly specified, at least ssse3

Vidi0:
Updated so the features are enabled by default

imLinguin (approved):

ChristopherHX (approved):
..., offline extract from 10s to 8s, this get's into nano optimizations I cannot notice at all...

ChristopherHX:
random fs noise could be bigger for me now 50% of what I noticed is indeed random noise

imLinguin:
It's possible the ARM codepath is faster by default

PR #61, opened 2026-07-01
fix: streaming check hash before writing

ChristopherHX:
Before this change every bit fault was written to disk after decryption

---

Drawback speed


ChristopherHX:
Breaks --try-skip-ntfs, we need CI tests

ChristopherHX:
performance slowdown very possible

PR #62, opened 2026-07-01
misc: reuse http client connection pool

ChristopherHX:
* draft due to mixed speedup / slowdown results hard to tell what is better


imLinguin:
re-using connection pool should be beneficial, alternatively having separate clients on long lived futures could also work

imLinguin:
From my testing it shows a big uplift in speed, as we are no longer making so many new TCP connections

ChristopherHX:
> no longer making so many new TCP connections

Yes this is true, now --parallel count connections are open the whole time.

Vidi0 (approved):

imLinguin (approved):

PR #66, opened 2026-07-03
Store CIKs in the keychain

Vidi0:
When a CIK (Content Integrity Key) is downloaded, store it in the keychain instead of forgetting it. Only try to fetch a CIK if it's not found in the keychain first.

Implement reading CIKs from the keychain for both local extract and streaming.

Also, now decrypting a XVD file that contains multiple encrypted regions with different keys is allowed (before it panicked with a todo).

### DIFF
--- a/xodus-cli/src/commands/license.rs
+++ b/xodus-cli/src/commands/license.rs
@@ -19,6 +19,7 @@ pub async fn run(
     tokio::fs::create_dir_all(&ciks).await.unwrap();
     for (uuid, content_key) in game_splicense.content_keys {
         let unpacked = content_key.unpack(&key).expect("failed to unpack");
+        tokens.save_cik(uuid, unpacked).unwrap();
         let mut file = OpenOptions::new()
             .create(true)
             .write(true)
--- a/msixvc/src/models/xvd/structs/parsed.rs
+++ b/msixvc/src/models/xvd/structs/parsed.rs
@@ -168,7 +168,7 @@ impl From<raw::XvdHashEntry> for XvdHashEntry {
 #[derive(Debug, Clone, PartialEq, Eq)]
 pub struct XvcInfo {
     pub content_id: Uuid,
-    pub xvc_encryption_key_id: HashMap<u16, Uuid>,
+    pub xvc_encryption_key_id: HashMap<u8, Uuid>,
     pub description: [u8; 0x100],
     pub version: u32,
     pub region_count: u32,
@@ -191,7 +191,7 @@ impl From<raw::XvcInfo> for XvcInfo {
                 .xvc_encryption_key_id
                 .into_iter()
                 .enumerate()
-                .map(|(i, uuid)| (i as u16, Uuid::from_bytes_le(uuid)))
+                .map(|(i, uuid)| (i as u8, Uuid::from_bytes_le(uuid)))
                 .filter(|(_i, id)| !id.is_nil())
                 .collect(),
             description: value.description,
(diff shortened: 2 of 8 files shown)


xodus-cli/src/commands/license.rs
```diff
@@ -19,6 +19,7 @@ pub async fn run(
     tokio::fs::create_dir_all(&ciks).await.unwrap();
     for (uuid, content_key) in game_splicense.content_keys {
         let unpacked = content_key.unpack(&key).expect("failed to unpack");
+        tokens.save_cik(uuid, unpacked).unwrap();
```
ChristopherHX, line 22: lease duration? for gamepass licenses

Vidi0: CIKs don't expire. It is just the two AES 128 keys joined together, it doesn't matter if you own the game in that moment, if you have the CIK for that package you can decrypt it.

Vidi0:
Forgot to mention, this allows to decrypt MSIXVC packages offline (if the CIK has already been fetched). Before, fetching Microsoft's servers was always required.

ChristopherHX:
IMHO we should store license not cik, a lot of XStore info is lost by this. e g. Trial and Full became the same.

imLinguin:
I don't know what to think about this. Launching will have to reach servers anyways eventually to support leases.
Some of the behavior of storing stuff locally is reserved to PC that has enabled the offline capability

Vidi0:
> IMHO we should store license not cik, a lot of XStore info is lost by this. e g. Trial and Full became the same.

The idea of this PR is that CIKs are static and don't expire, so IMO it makes sense to store them in the keychain and reuse them. More information could also be stored, but for decryption CIKs are the only thing needed, so that's what I choose to store. If more information is required for launching, then that could be added later, but right now it would only add complexity.

Vidi0:
> I don't know what to think about this. Launching will have to reach servers anyways eventually to support leases. Some of the behavior of storing stuff locally is reserved to PC that has enabled the offline capability

AFAIK, this is not exclusive to PC. CIKs are static, if the package is encrypted with a given CIK, it can be decypted with that key at any time. This change is not related to launching, just extraction/decryption, so I don't think I understand what you're trying to tell me there. Yes, probably launching will require a connection, but for extraction it's nice to be able to decrypt it offline.

The main reason I made this PR is because when I was trying to benchmark the extraction code, we had reached the point where those network requests were actually costing ~30% of the time, so it was messing with the results.

ChristopherHX:
> The main reason I made this PR is because when I was trying to benchmark the extraction code, we had reached the point where those network requests were actually costing ~30% of the time, so it was messing with the results.

What I did once locally for streaming test, just added local test code to skip the http.

> The idea of this PR is that CIKs are static and don't expire

Technically this is true, but legally not always. I did intentionally do the license request via http until the legal cik lifetime (not technical cik lifetime) are sorted out.
You shift the responsibility of cik lifetime to xodus, which is legally not always static.

e.g.
- only the primary pc is allowed to download/extract/run the game offline (the download probably not offline, but nevertheless)
- the primary pc can only be changed n times per year


I belive my windows 11 device has lost it's primary pc bit due to me exterimenting with xodus/win11 vms already. Now refuses to run it offline, I bet it also has no access to the cik from any persistent media

ChristopherHX:
with fuse, we could just extract encrypted and mount it unencrypted or whatsoever

Vidi0:
> Technically this is true, but legally not always. I did intentionally do the license request via http until the legal cik lifetime (not technical cik lifetime) are sorted out.

I think the lifetime of the CIK is not an issue. First of all, this does not matter for owned games, as you get the CIK when you buy it and you have a permanent license for it. For that reason, this would only matter for GamePass games.

Then, for GamePass, if Microsoft allows you to get a CIK it means that they allow you to decrypt a game. When you obtain a CIK, you could think of it as if you had also downloaded the encrypted package, decrypted it and stored it in a hard drive. This PR only helps avoid the technical limitation of having to store all the decrypted games, as now storing a CIK is the only thing needed.

The checks that verify that you own the game should be performed IMO at the time of launch.

PR #68, opened 2026-07-07
Add Trial variant to DeviceCredential enum

yousif51811:
It allows you to install games that have trial variants.


imLinguin:
Right now we will hold off on that variant mostly because we can't yet tell games their license type through SDK, nor we are yet enforcing the actual ownership type

PR #69, opened 2026-07-07
Remove `read_struct!` macro

Vidi0:
I don't know why I didn't think of this before... If we're using a `impl_struct!` macro, just implement the functions `read` and `from_array` directly for each struct. We couldn't take advantage of the trait anyways because we needed to use a macro to read.

Also, remove a unused trait.


imLinguin (approved):
I like this

PR #70, opened 2026-07-08
Add manual GPT parse

Vidi0:
Replace the parse of the GPT partition table with our own implementation, which is Async and does not require Seek (useful for streaming, as it avoids network requests). The idea would be to replace the NTFS partition parsing, too, but that is a lot more complex.

I'm pushing this as a proof of concept, as I don't really like how it turned out. It feels fragile to me: I intended to do a very simple parsing knowing how the partition tables of XVCs are laid out in practice, but ended up making it too complex, capable of parsing any valid GPT partition table. Still, it doesn't cover every possible invalid header, which makes this implementation an odd middle point between simple parsing and complete/robust parsing.

It works, though.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -875,6 +875,15 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "217698eaf96b4a3f0bc4f3662aaa55bdf913cd54d7204591faa790070c6d0853"
 
 [[package]]
+name = "crc32fast"
+version = "1.5.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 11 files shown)


imLinguin:
Could you point to some documentation on GPT? I'd like to do some reading, before looking deeper into this code

Vidi0:
I just used the wikipedia page:
https://en.wikipedia.org/wiki/GUID_Partition_Table

BellezaEmporium:
> I just used the wikipedia page: https://en.wikipedia.org/wiki/GUID_Partition_Table

While it is complete, there's also a MS Learn FAQ that goes into other details on what's going on : https://learn.microsoft.com/en-us/troubleshoot/windows-server/backup-and-storage/guid-partitioning-table-disk-faq

PR #71, opened 2026-07-10
feat: support decrypting sign and hmac clep blobs

imLinguin:
this PR aims to add TpmInfo data to deviceaddcredential.srf call, which appears to be required to do proper Xbox device authentication


xodus/src/licensing/utils.rs
```diff
@@ -7,3 +9,50 @@ pub fn generate_suid() -> String {
...
+            let _dq = take(cb_p2);
+            let _qinv = take(cb_p1);
+            let d = take(cb_mod);
+            RsaPrivateKey::from_components(n, e, d, vec![p, q])
+        }
+        RSAPRIVATE_MAGIC => {
+            // No d in the blob — recompute it.
+            let one = BigUint::from(1u32);
+            let lambda = (&p - &one) * (&q - &one); // or lcm for stricter form
+            let d = BigUint::from_bytes_be(&e.to_bytes_be())
```
BellezaEmporium: Adding "to_bytes_be" seems redundant to me. e is technically already BigUint from the take command. You may directly go with .mod_inverse(&lambda)

xodus/src/licensing/splicense.rs
```diff
@@ -296,37 +323,79 @@ impl SPLicense {
...
-    pub fn parse_base64(string: String) -> Result<SPLicense, SPLicenseParseError> {
+    pub fn parse_base64(string: &str) -> Result<SPLicense, SPLicenseParseError> {
         let data = BASE64_STANDARD.decode(string)?;
         Ok(SPLicense::decode(&*data)?)
     }
 }
 
-impl EncryptedDeviceKey {
-    fn decryption_key(&self) -> [u8; 16] {
+pub trait EncryptedBuffer {
```
Vidi0: I wouldn't use a Trait here. I would prefer having a free function:
```
fn decryption_key(key_schedule: [u32; 58]) -> [u8; 16] {
```
And then (if really needed, if it's only going to be used privately it's not needed).
```
impl T {
  pub fn decryption_key(&self) ->[u8; 16] {
    decryption_key(self.key_schedule)
  }
}
```

xodus/src/licensing/splicense.rs
```diff
@@ -296,37 +323,79 @@ impl SPLicense {
...
+        key[1] = key_schedule[36] ^ key_schedule[47] ^ 0xDF080E39;
+        key[2] = key_schedule[40] ^ key_schedule[51] ^ 0x6D09B2F5 ^ 0x2AE17AB9;
+        key[3] = key_schedule[30] ^ key_schedule[41] ^ 0x37288CEC;
 
         transmute!(key)
     }
 
+    /// Decrypts `data` with AES-128-CBC (zero IV) using the key derived from `key_schedule`.
+    fn decrypt_cbc_zero_iv<const N: usize>(key_schedule: [u32; 58], data: &[u8; N]) -> [u8; N] {
+        let key = Self::decryption_key(key_schedule);
```
Vidi0: If data must have a length multiple of 16, assert it at compile time.

```suggestion
        const { assert!(N.is_multiple_of(16)) }
                
        let key = Self::decryption_key(key_schedule);
```

xodus/src/licensing/splicense.rs
```diff
@@ -296,37 +323,79 @@ impl SPLicense {
...
+        let aes = aes::Aes128::new(&key.into());
+
+        let mut out = [0u8; N];
+        let mut prev: u128 = 0;
+        for (chunk_in, chunk_out) in data.chunks_exact(16).zip(out.chunks_exact_mut(16)) {
+            let block: [u8; 16] = chunk_in.try_into().unwrap();
+            let next: u128 = transmute!(block);
```
Vidi0: This dosen't take the endianness of the machine into account. It's better to use the following:

```suggestion
            let next = u128::from_le_bytes(block);
```

xodus/src/licensing/splicense.rs
```diff
@@ -296,37 +323,79 @@ impl SPLicense {
...
+
+            let mut decrypted = block.into();
+            aes.decrypt_block(&mut decrypted);
+            let decrypted: [u8; 16] = (*decrypted).try_into().unwrap();
+            let decrypted: u128 = transmute!(decrypted);
```
Vidi0: Same as above:

```suggestion
            let decrypted = decrypted.to_le_bytes();
```

xodus/src/licensing/splicense.rs
```diff
@@ -296,37 +323,79 @@ impl SPLicense {
...
     }
 
+    /// Decrypts `data` with AES-128-CBC (zero IV) using the key derived from `key_schedule`.
+    fn decrypt_cbc_zero_iv<const N: usize>(key_schedule: [u32; 58], data: &[u8; N]) -> [u8; N] {
+        let key = Self::decryption_key(key_schedule);
+        let aes = aes::Aes128::new(&key.into());
+
+        let mut out = [0u8; N];
+        let mut prev: u128 = 0;
+        for (chunk_in, chunk_out) in data.chunks_exact(16).zip(out.chunks_exact_mut(16)) {
```
Vidi0: Instead of chunks exact and then try into array, we can use the const generic version: `as_chunks` (and `as_chunks_mut`), which returns an slice over arrays directly:

```suggestion
        for (chunk_in, chunk_out) in data.as_chunks::<16>.0.iter().zip(out.as_chunks_mut::<16>) {
```

xodus/src/licensing/splicense.rs
```diff
@@ -296,37 +323,79 @@ impl SPLicense {
...
+            let block: [u8; 16] = chunk_in.try_into().unwrap();
+            let next: u128 = transmute!(block);
+
+            let mut decrypted = block.into();
+            aes.decrypt_block(&mut decrypted);
+            let decrypted: [u8; 16] = (*decrypted).try_into().unwrap();
```
Vidi0: I think you can access the inner array directly.

```suggestion
            let decrypted = decrypted.0;
```

xodus/Cargo.toml
```diff
@@ -12,23 +12,26 @@ serde_json = "1.0.145"
 quick-xml = { version = "0.40.1", features = ["serialize", "overlapped-lists"] }
 log = "0.4.28"
 chrono = { workspace = true }
-base64 = "0.22.1"
+base64 = { workspace = true }
 rand = "0.10.0"
+rand08 = { version = "0.8", package = "rand" }
```
Vidi0: Why is this needed? Why not use the latest version?

xodus/src/clep/challenge.rs
```diff
@@ -44,31 +44,108 @@ impl Cipher {
 
     const INITIAL_STATE: u32 = Cipher::initial_state();
 
+    // The 8 round functions shared by the key schedule and the Feistel
+    // cipher below: the key schedule walks them forward (round_1..round_8),
+    // the cipher walks the same functions backward (round_8..round_1).
+    const fn round_1(x: u32) -> u32 {
```
Vidi0, line 51: I would probably mark all rounds with the inline attribute: `#[inline]` (just to make sure, the compiler should do it automatically).

Vidi0 (dismissed):

xodus/src/clep/challenge.rs
```diff
@@ -44,31 +44,108 @@ impl Cipher {
...
+            .wrapping_sub(x.rotate_right(9))
+    }
+
+    const fn whiten(x: u32) -> u32 {
+        Self::MAGIC_03
+            .wrapping_mul(x.wrapping_add(Self::MAGIC_HI).rotate_right(10))
+            .wrapping_sub(x.rotate_right(29))
+    }
+
+    const fn round(id: u8, x: u32) -> u32 {
```
Vidi0: This is odd 🤔. Why not use the rounds directly?

xodus/src/clep/challenge.rs
```diff
@@ -44,31 +44,108 @@ impl Cipher {
...
+        }
+        (a, b)
+    }
+
+    /// Inverse of `run_rounds`: given the state produced by
+    /// `run_rounds(a, b, rounds)`, recovers the original `(a, b)`.
+    ///
+    /// Each forward step `(a, b) -> (b, a ^ round(b))` is undone by
+    /// `(a, b) -> (b ^ round(a), a)`, walking `rounds` back to front.
+    const fn run_rounds_inverse(mut a: u32, mut b: u32, rounds: [u8; 8]) -> (u32, u32) {
```
Vidi0: This is just `run_rounds` with inverse order, right? `run_rounds` should be used instead, and this function removed. Or, even better remove both `run_rounds` and `run_rounds_inverse` and add:

```
const fn run_rounds(a: u32, b: u32) -> (u32, u32) {
    let (a, b) = (b, a ^ Self::round_1(b));
    let (a, b) = (b, a ^ Self::round_2(b));
    let (a, b) = (b, a ^ Self::round_3(b));
    let (a, b) = (b, a ^ Self::round_4(b));
    let (a, b) = (b, a ^ Self::round_5(b));
    let (a, b) = (b, a ^ Self::round_6(b));
    let (a, b) = (b, a ^ Self::round_7(b));
    let (a, b) = (b, a ^ Self::round_8(b));
    (a, b)
}

const fn run_rounds_inverse(a: u32, b: u32) -> (u32, u32) {
    let (a, b) = (b ^ Self::round_8(a), a);
    let (a, b) = (b ^ Self::round_7(a), a);
    let (a, b) = (b ^ Self::round_6(a), a);
    let (a, b) = (b ^ Self::round_5(a), a);
    let (a, b) = (b ^ Self::round_4(a), a);
    let (a, b) = (b ^ Self::round_3(a), a);
    let (a, b) = (b ^ Self::round_2(a), a);
    let (a, b) = (b ^ Self::round_1(a), a);
    (a, b)
}
```

Then `Cipher::round` could be removed as well.

Other idea could be having a struct that holds both `a` and `b` and has a `const fn round_*(&mut self)` which modifies `self`. Then instead of passing tuples around, we pass that struct. `Cipher` would then be `plain: u64` + that struct.

ChristopherHX (approved):
Working xbox live device auth RPC ticket 🎉

PR #72, opened 2026-07-14
feat: run command

### LINKED ISSUE #33: Have a nice little progressbar for extract
So you have an indicator and an estimated time when the game will be ready to go.

### LINKED ISSUE #50: Keep files encrypted on disk
As a point of security against easy piracy, encrypted game files should remain so when extracted. Using [containers/fuse-overlayfs](https://github.com/containers/fuse-overlayfs) we should be able to pass the decrypted content to WineGDK or GDK-Proton with umu-launcher.

This would require another command line option, I suggest:
```sh
xodus-cli run /path/to/encrypted/installation/dir [proton (default) / wine]
```
We should accept a path to wine / proton in either a config file and/or an environment variable (which would override the config file)

ChristopherHX:
* Works best with this ntdll.so patch https://github.com/ChristopherHX/WineGDK/commit/4119fd28a6efb70383df29d3af5061074165caef (only a single commit to cherry-pick)
* Extract is now based on streaming

Closes #33
Closes #50


ChristopherHX:
Example usage, you can also use the streaming extract command
```
cargo run -r -- streaming 9NBLGGH2JHXJ ../out-streaming2
WINEPREFIX=~/.wine_tmp_2 cargo run -r -- run ../out-streaming2 ~/Documents/minecraft/WineGDK/wine
```

ChristopherHX:
...I find often some logic bugs here and there...

ChristopherHX:
should still add hdi tools for macOS..., taskmanager shows writes to disk via mkstemp

crates/xodus-cli/src/commands/run.rs
```diff
@@ -0,0 +1,165 @@
...
+    source: String,
+    wine: String,
+    exe: Option<String>,
+    market: Option<String>,
+) {
+    let mut lfiles: HashMap<String, SegmentFile> = HashMap::new();
+
+    let out: &Path = Path::new(&source);
+    let out_absolute = std::fs::canonicalize(out).unwrap();
+    let final_path = out.join(".xodus-streaming.msixvc");
```
ChristopherHX, line 130: Maybe we can test out if we put the ../uuid in here if we can launch a windows file from cold ntfs

imLinguin (approved):
Let's do this

PR #74, opened 2026-07-18
msixvc: add RegionDecryptor

Vidi0:
The new `DecryptorReader` struct reads bytes from an underlying reader in chunks of `PAGE_SIZE` bytes, decrypts each page from the provided `RegionTable` decryption info, and implements `Read`, `ReadBuf`, `Seek` in order to access the decrypted data.

By avoiding doing the AES key expansion and all that stuff on each read, this should be faster than the current `XvdStream` that creates a new `SectionReader` on each read.

The idea was that the `RegionTable` was stored inside `XvdFile` (created inside `XvdFile::new()`), computed once, and then whenever we need to read the data inside the MSIXVC, we just create a new `DecryptorReader` once, and reuse it for every read.

Also, the `DecryptorReader` (and `RegionDecryptor`, which does the decryption itself) is generic over `Units`, so we can use `Vec<u32>` now, and then move over to using `Arc<[u32]>` if we want to do async parallel decryption. With `Arc` we could cheaply clone the `RegionDecryptor` (and thus the `RegionTable` which contains them), and send one decryptor to each async thread.

However, I did not implement any of that, because the code in `xvd.rs` is very complex and I couldn't figure out how to add the `DecryptorReader` into there. I would like to refactor that file before integrating the decryptor. Still, I would like to merge this new decryptor so new code can depend on it directly instead of the old `SectionReader`, which has many performance issues like requiring a seek for each read and not being able to span multiple regions in one single reader.


Vidi0:
Sorry for the amount of commits, I usually try to keep the number of commits small in order to make it easier to review one by one, but in this case I had to do many refactors until I ended up with this design, so just review the changes directly.

ChristopherHX:
Meanwhile me already propose deletion of `SectionReader` as I hope to have preannounched #72 

Having a newer better interface that has no reader / writer attached was in my mind for todo.

Need to look at your concept

Vidi0:
In this implementation `RegionDecryptor` could be used directly if we need to decrypt without `Read` or `ReadBuf`. However, the `DecryptorReader` is a lot more ergonomic, reading in chunks automatically and outputting bytes as they are read.

I'll have to look at your idea, too.

msixvc/src/crypt/region.rs
```diff
@@ -0,0 +1,147 @@
+use super::RegionDecryptor;
+use crate::models::xvd::PAGE_SIZE;
+
+use std::cmp::Ordering;
+use std::range::Range;
```
ChristopherHX, line 5: use of unstable library feature `new_range_api`
see issue #125687 <https://github.com/rust-lang/rust/issues/125687> for more informationrustc[Click for full compiler diagnostic](rust-analyzer-diagnostics-view:/diagnostic%20message%20[0]?0#file:///Users/christopher/Documents/minecraft/xodus/msixvc/src/crypt/region.rs)

ChristopherHX (changes requested):
Toolchain minimum bump, needs careful consideration.

BREAKING change for my local setup

msixvc/src/crypt/region.rs
Vidi0: The feature is stable, it should be available in the latest Rust version.

ChristopherHX: Readme mentions 2024 rust edition, but not that it needs to be from June/July.

At least update the readme then I update

PR #76, opened 2026-07-19
feat: MSATokenResponse return device_rps if possible

ChristopherHX:


imLinguin (approved):

ChristopherHX:
user token does not work, but that is not caused here

PR #77, opened 2026-07-19
docs: update docs with clep and new device token findings

imLinguin:


.gitignore
```diff
@@ -1,5 +1,2 @@
 /target
-.env
```
ChristopherHX, line 2: `.env` does not look to be that useless for me

imLinguin: Xodus never loaded a .env so I thought it was redundant, I can add it back

docs/xodus/device.md
```diff
@@ -62,17 +62,23 @@ All blobs are base64 encoded
...
+
 ## RST2.srf
 
-In below example it is possible to ommit signatures in this requiest in order to receive raw unencrypted data
+Omitting the `<Signature>` element from this request causes the server to return raw, unencrypted data instead of an encrypted payload - useful for inspecting responses while reverse engineering, but not something a real client does.
+
+The first `RST2.srf` call after provisioning exchanges the device credential for a device STS token. Only `Username` is sent in `wsse:UsernameToken` - the password from `deviceaddcredential.srf` would also be accepted here, but a password-authenticated device token isn't trusted enough to be used against Xbox services. Instead, the whole request is signed (`rsa-sha256`) with the device's private RSA key, which is recovered by decrypting `ClepSignState` from the `SPLicenseBlock` returned by `deviceaddcredential.srf`. See [CLEP secrets](./clep.md#clepsignstate) for how that key is extracted.
```
ChristopherHX, line 81: The variant with password uses a different algorithm id.

e.g. the secure version has DOUBLEDERIEVED, I would add that here

ChristopherHX: TPMInfo alters the RST2 response with rsa signing

e.g. Creates an EncryptedKey object that changes the hmac encryption

imLinguin: Im not sure about the DOUBLEDERIVED part, as that was always the value we sent in the SOAP envelope, even when using Password

.gitignore
ChristopherHX: I place path to protoc there, which we barely use in service atm.

docs/xodus/device.md
ChristopherHX: `<wssc:DerivedKeyToken wsu:Id="SignKey" Algorithm="urn:liveid:SP800-108CTR-HMAC-SHA256">`

e.g. if you just need to rewrite the device auth to pwd and windows and service responses use.

I have traces of this behavior from windows and service

Which makes sense, since the clep derivation is missing

Normally windows always logs `Algorithm="urn:liveid:SP800108_CTR_HMAC_SHA256_DOUBLEDERIVED"`, since it uses the clep rsa flow.

ChristopherHX (approved):

PR #78, opened 2026-07-20
fix: clientid= not working for user token

ChristopherHX:


ChristopherHX:
does only work for full trust :(

imLinguin:
I've seen `clientId` in some traces, could it matter here?

imLinguin:
Closed in favor of #86

PR #82, opened 2026-07-20
Example Borrow cow

ChristopherHX:
* incomplete rewrite with lifetime qualifier

### DIFF
--- a/xodus/src/models/soap/crypto.rs
+++ b/xodus/src/models/soap/crypto.rs
@@ -1,4 +1,5 @@
 use serde::{Deserialize, Serialize};
+use std::borrow::Cow;
 
 use super::base::ReferenceUri;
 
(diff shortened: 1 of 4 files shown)


xodus/src/models/soap/crypto.rs
```diff
@@ -59,64 +60,64 @@ impl SignatureKeyInfo {
...
 
-impl SignatureReference {
+impl<'t> SignatureReference<'t> {
     pub fn exclusive(uri: &str) -> Self {
         Self {
             uri: uri.to_string(),
             transforms: SignatureTransforms {
                 transform: vec![AlgorithmNode {
-                    algorithm: "http://www.w3.org/2001/10/xml-exc-c14n#".to_string(),
+                    algorithm: Cow::Borrowed("http://www.w3.org/2001/10/xml-exc-c14n#"),
```
ChristopherHX, line 92: should not clone the data here, not sure what into would to instead. I got lost.

PR #84, opened 2026-07-21
feat: add RSTRequestBuilder

### LINKED ISSUE #80: [Tracking] Refactor RST2.srf token logic
`exchange_device_token` and `exchange_user_token` are becoming quite big. Given how configurable those endpoints, I think it's time to:
- Introduce a `RSTRequestBuilder`
  - Builder should accept all key material needed to sign the request with `rsa-sha256` and `hmac-sha256`
  - Builder should allow for adjusting AuthInfo structure when it comes to
    - hostingApp
    - InlineFT
    - InlineUX
  - The structure should be re-useable across device and user tokens
- Introduce proper error handling
  - As part of the effort of removing all unwrap and expects from core, we should be able to handle recoverable errors on user tokens, but also properly return faults to callers

imLinguin:
I also took some liberty to include changes to adding TPMInfo again, but the OAEP decryption isn't added yet

### DIFF
--- a/xodus-service/src/connection/xml.rs
+++ b/xodus-service/src/connection/xml.rs
@@ -78,10 +78,9 @@ pub async fn parse_message(
 
             let user_token = xodus::api::live::exchange_user_token(
                 &context.client,
-                token.token,
+                token,
                 "USERNAME".to_string(),
-                device_token.token.clone(),
-                device_token.binary_secret.clone().unwrap(),
+                device_token.clone(),
                 None,
                 Some("Silent".to_string()),
                 req.client_id.clone(),
--- a/xodus-service/src/connection/xml.rs
+++ b/xodus-service/src/connection/xml.rs
@@ -53,14 +53,11 @@ pub async fn parse_message(
             };
             let (scope, client_id) = if req.msa_full_trust {
                 (
-                    "service::user.auth.xboxlive.com::MBI_SSL".to_owned(),
+                    "service::user.auth.xboxlive.com::MBI_SSL",
                     format!("clientid={}", req.client_id),
                 )
             } else {
-                (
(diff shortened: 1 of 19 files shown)


xodus/Cargo.toml
```diff
@@ -14,6 +14,7 @@ log = "0.4.28"
 chrono = { workspace = true }
 base64 = { workspace = true }
 rand = "0.10.0"
+rand08 = { version = "0.8", package = "rand" }
```
imLinguin, line 17: This version is used by rsa, this seemed to be the only way to get rng for rsa, let me know if you have alternatives

ChristopherHX:
> OAEP decryption isn't added yet

I am not 100% sure that this magically solves user auth, since my windows 10 vm didn't had sign in problems without tpminfo (but I had to disable the proxy time to time due to skill issues with connectivity)

xodus-service/src/connection/xml.rs
```diff
@@ -51,10 +51,16 @@ pub async fn parse_message(
             let Token::Legacy(token) = context.tokens().get_user_sts_token()? else {
                 return Ok(vec![]);
             };
-            let scope = if req.msa_full_trust {
-                "service::user.auth.xboxlive.com::MBI_SSL".to_owned()
+            let (scope, client_id) = if req.msa_full_trust {
+                (
+                    "service::user.auth.xboxlive.com::MBI_SSL".to_owned(),
```
ChristopherHX: string literals have type &'static str, so to_owned() is only creating an heap copy for nothing

ChristopherHX (approved):
Looks good to me, only login command has not been tested by me.

imLinguin:
I have tested both device + user logins, all is correct.  
Licensing also works

PR #87, opened 2026-07-22
fix: return error result for decrypt_response, not panic

ChristopherHX:
* only kept Options unwrap for now


imLinguin (approved):
Good change, in the end #84 should remove this function in favor to better typed error type buit with thiserror

ChristopherHX:
> typed error type buit with thiserror

What I miss are stacktraces, to the origin...., I have issues to follow the origin of loose errors..

PR #97, opened 2026-07-23
refactor: apply clippy suggestions

imLinguin:
- enum naming convention
- use box in enum to avoid size inflation
- prefer String::new() instead of "".to_string()
- use as_mut when modifying header Option type in RSTRequestBuilder

additionally I added a few fixes from pedantic ruleset

### DIFF
--- a/xodus/src/api/live/utils.rs
+++ b/xodus/src/api/live/utils.rs
@@ -95,7 +95,7 @@ pub fn sign_xml(
 }
 
 pub fn decrypt_soap_encrypted_data<T: serde::de::DeserializeOwned>(
-    encrypted_data: soap::EncryptedData,
+    encrypted_data: Box<soap::EncryptedData>,
     signature: &rst::RSTSignature,
     nonces: &HashMap<String, String>,
 ) -> Result<T, rst::RSTError> {
--- a/xodus/src/tokens/device.rs
+++ b/xodus/src/tokens/device.rs
@@ -74,7 +74,7 @@ async fn reauthenticate_device(client: &reqwest::Client, tokens: &TokenManager, 
 
 fn save_device_sts_token(
     tokens: &TokenManager,
-    resp: crate::models::soap::RequestSecurityTokenResponse,
+    resp: Box<crate::models::soap::RequestSecurityTokenResponse>,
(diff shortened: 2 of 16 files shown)


xodus/src/tokens/device.rs
```diff
@@ -74,7 +74,7 @@ async fn reauthenticate_device(client: &reqwest::Client, tokens: &TokenManager,
 
 fn save_device_sts_token(
     tokens: &TokenManager,
-    resp: crate::models::soap::RequestSecurityTokenResponse,
+    resp: Box<crate::models::soap::RequestSecurityTokenResponse>,
```
ChristopherHX, line 77: I would prefer borrowing here, instead of boxing

xodus/src/api/live/utils.rs
```diff
@@ -95,7 +95,7 @@ pub fn sign_xml(
 }
 
 pub fn decrypt_soap_encrypted_data<T: serde::de::DeserializeOwned>(
-    encrypted_data: soap::EncryptedData,
+    encrypted_data: Box<soap::EncryptedData>,
```
ChristopherHX, line 98: Same here I prefer borrowing

ChristopherHX:
> prefer String::new() instead of "".to_string()

possible an implementation detail

xodus/src/tokens/device.rs
imLinguin: I made this a Box since the enum itself provides a Box now - in order to reduce the max size of it on stack.
Given that the value is not needed after this point I just pass its ownership to this function

xodus/src/api/live/utils.rs
imLinguin: Same as I mentioned above, the goal of this was to pass the ownership, the value is dropped here. that field was already boxed when coming from serde - to reduce the stack size of enum

ChristopherHX (approved):
the thing is if it isn't in a Box you need to heap allocate it, but the .into usage is a part that is possible a problem.... Would we only have as_token

imLinguin:
We should really try to borrow as much as possible when serializing. I'll try to explore that with your drafts

PR #99, opened 2026-07-24
misc: add claude files

imLinguin:
this PR adds basic CLAUDE.md, skills and settings  
the goal is to help accelerate analysis of network traffic and documenting any findings.

AI could also be used for prototyping of new features, but I will strongly push against shipping full features with AI without understanding new code and not being confident about its quality


imLinguin:
Scratching this for now

PR #101, opened 2026-07-25
fix: encrypted file streaming check

ChristopherHX:
* prevent remaining to go less than zero


imLinguin (approved):
ok, that bug is a quite ridiculous

ChristopherHX:
we need ci test for this..., at best with self signed msixvc files

ChristopherHX:
the last tests I did was with offline extract

imLinguin:
We could have CI tests with our own packages, but I think well known key is what we are after, one we can package with makepkg tool

PR #107, opened 2026-08-01
fix: return the device token do_sisu was authorized with

AllanVester:
`do_sisu` mints a device token via `get_device_token_rps` and then drops it. The title token SISU returns is bound to that device, so a caller that wants an XSTS from the SISU bundle has no way to supply a matching device token - minting a fresh one via `get_device_token()` uses `XalAuthenticator::device_id = Uuid::new_v4()`, a different machine every instance, and `xsts.auth` rejects the pair with `401 XSTS error="title_usage_by_device_exceeded"` (XErr 2148916254).

Return it as a third tuple element so the device token, the title token bound to it, and the user token can all describe one device.

### DIFF
--- a/crates/xodus/src/auth.rs
+++ b/crates/xodus/src/auth.rs
@@ -77,6 +77,7 @@ pub async fn do_sisu(
     (
         XalAuthenticator,
         xal::response::SisuRPSAuthorizationResponse,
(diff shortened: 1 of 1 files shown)


ChristopherHX (approved):
Yes a very good idea to return it as well

PR #108, opened 2026-08-01
feat: mint title-claimed XSTS tokens over IPC

AllanVester:
Adds XSTS_TOKEN_REQUEST/RESPONSE (5/6) and a handler that runs SISU for the caller's client id and title id, then exchanges the bundle for an XSTS scoped to one relying party.

A caller cannot do this itself. The device token, the SISU title token bound to it, and the user token must all describe one device, and only this side holds the device identity - so only this side can keep the three consistent. The observable difference is the userhash: a token carrying a TITLE claim gets a different uhs per relying party, while a token minted from a user token alone gets one uniform uhs everywhere.

The response carries the uhs and the XUID alongside the token. The uhs is the `x=` half of `XBL3.0 x=<uhs>;<token>` and is per-token. The XUID is needed because callers that mint locally read `xid` out of their own XSTS response; a token minted here must hand it back, or the caller's user id silently stays 0 and its requests go out as `users/xuid(0)/...` and are refused.

### DIFF
--- a/crates/xodus-service/src/connection/xml.rs
+++ b/crates/xodus-service/src/connection/xml.rs
@@ -4,7 +4,9 @@ use xodus::{
         live::ExchangeUserTokenOutcome,
         secrets::Token,
         soap,
-        xgameruntime::xuser::{MSATokenRequest, MSATokenResponse},
+        xgameruntime::xuser::{
+            MSATokenRequest, MSATokenResponse, XstsTokenRequest, XstsTokenResponse,
+        },
     },
     proto::xodus::XodusMessageType,
 };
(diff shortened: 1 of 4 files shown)


olivi-r:
This can be done by the client with the existing `MSA_TOKEN_RESPONSE`
```
POST https://device.auth.xboxlive.com/device/authenticate
Signature: ...
{
  "RelyingParty": "http://auth.xboxlive.com",
  "TokenType": "JWT",
  "Properties": {
    "AuthMethod": "RPS",
    "ProofKey": <signature proof jwt>,
    "RpsTicket": <device rps ticket>,
    "SiteName": "user.auth.xboxlive.com"
  }
}
``` 
```
POST https://sisu.xboxlive.com/authorize
Signature: ...
{
  "AccessToken": <user rps ticket>,
  "DeviceToken": <device token from previous request>,
  "Sandbox": "RETAIL"
}
```

ChristopherHX:
If we add sisu token caching I like the direction, giving the xgameruntime so much power is something I think is at the moment ok however should change long term.

olivi-r:
Possibly, but keeping the interaction to only a minimal request on login keeps xodus stateless across different wine/proton instances, otherwise signing keys would need to be tracked and an extra ipc call made for each call to XUserGetTokenAndSignature.

Issue #109, opened 2026-08-03
Implement device - user association


imLinguin:
Right now we do not implement anything to display the device on users account to let them have control over revoking the token, as well as reserving the spot as one of 10 possible devices at the same time.

We should implement the DeviceAssoc.srf endpoint after user login.
Not only would it provide the above, but also potentially help with primary device configuration support.

PR #110, opened 2026-08-03
feat: logout cli command

olivi-r:
Add a `logout` command that removes stored persistent data.

### DIFF
--- a/crates/xodus-cli/src/commands.rs
+++ b/crates/xodus-cli/src/commands.rs
@@ -3,6 +3,7 @@ pub mod download;
 pub mod extract;
 pub mod license;
 pub mod login;
+pub mod logout;
 #[cfg(unix)]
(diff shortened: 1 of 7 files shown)


olivi-r:
Once #109 is implemented, this could be enhanced to remove the device association.

imLinguin:
We should not remove device token by default, but add a flag to do so.

ChristopherHX:
> We should not remove device token by default

but what if it is from an ancient xodus build using now unsupported less secure device token format? then this makes logout alone less usable as a trivial fix.

imLinguin:
We should add a flag like --with-device 
If we supported offline configs accidental logouts could prevent from modifying that setting for a year, hence I think we should expect this behavior from the start.

imLinguin (approved):

PR #111, opened 2026-08-05
feat: key-chain-file for development

ChristopherHX:
* for testing purposes e.g. avoid keypass prompt on mac on rebuild
* cargo build --features key-chain-file
* allow share login between os and wine etc.

### DIFF
--- a/crates/xodus/Cargo.toml
+++ b/crates/xodus/Cargo.toml
@@ -49,3 +49,6 @@ ed25519-dalek = "2.2.0"
 
 [build-dependencies]
 prost-build = "0.14.4"
+
+[features]
+key-chain-file = []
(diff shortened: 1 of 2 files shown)


imLinguin (approved):
for testing writing to HOME is fine
in ideal scenarios XDG_DATA_HOME and similar data directories could be used too

Vidi0:
I feel like this should be togglable at run time, not compile time. Maybe a bool in clap:
```
--keychain-backend {file, keyring}
```

imLinguin:
I dont think such toggle makes sense, especially given that the file is plaintext

ChristopherHX:
xodus currently does not allow gracefully changing secret storage backends, so having it at runtime might break users?

PR #112, opened 2026-08-05
tech: add rust analyzer config

imLinguin:
adding enforced import granularity will improve readability quite a bit


ChristopherHX (approved):
So rust-analyze import fixer now maximally generates
```
use tokio::fs::{File, OpenOptions};
use tokio::io::AsyncRead;
use tokio::sync::mpsc::{Receiver, Sender};
```

imLinguin:
Yes, I think this is desired behavior.

PR #114, opened 2026-08-06
deps : bump, fix functions related to changes

BellezaEmporium:
# What ?

This PR updates the dependencies used by Xodus. Breaking changes include bergshamra v0.8 (incl. bergshamra-c14n) and reqwest v0.13.

# Why ?

To keep the project up to date, and prevent us from any security issues implied from older versions of concerned dependencies.

# Breaking changes

Bergshamra v0.8 includes a full API restructuring, breaking the current Hmac & Rsa usage that was done before. This PR fixes it. It also adds an external crate dependency known as kryptering.
The inclusion of reqwest v0.13 shatters out with XAL's 0.11. The "add_cv()" function that's used in "get_licence_content" cannot be used, and thus, needs to be added manually. A comment has been added to reflect that small regression.

# Regressions

-> Rsa:BigUint & num_bigint_dig do not like each other. To fix that, a "severe" requalification of the different variables was done in utils.rs. Might be subject to changes, based on how it is dealt at a later stage of the project.

-> add_cv() does not work with RequestBuilder. xal-rs (external crate, called for the function) uses an older version, which might be the reason why it doesn't properly work right now.

# Compilation result

```
shadix@TUFG-ISMO:/mnt/c/Users/*****/xodus-acdevs$ cargo build
   Compiling xodus v0.1.0 (/mnt/c/Users/*****/xodus-acdevs/crates/xodus)
   Compiling xodus-service v0.1.0 (/mnt/c/Users/*****/xodus-acdevs/crates/xodus-service)
   Compiling xodus-cli v0.1.0 (/mnt/c/Users/*****/xodus-acdevs/crates/xodus-cli)
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 1m 23s
```

### DIFF
--- a/crates/xodus/Cargo.toml
+++ b/crates/xodus/Cargo.toml
@@ -4,37 +4,38 @@ version = "0.1.0"
 edition = "2024"
 
 [dependencies]
-xal = { git = "https://github.com/xodus-gaming/xal-rs", rev = "b96a271aa216b78ebcf32dc67e0f712792b3e058", version = "0.1.3" }
+xal = { git = "https://github.com/xodus-gaming/xal-rs", rev = "6aa2420f8f0d5e6f7460ff13f399037583a942f4", version = "0.1.3" }
 tokio = { workspace = true }
 reqwest = { workspace = true }
-serde = { version = "1.0.228", features = ["derive"] }
-serde_json = "1.0.145"
-quick-xml = { version = "0.40.1", features = ["serialize", "overlapped-lists"] }
+serde = { version = "1.0.229", features = ["derive"] }
+serde_json = "1.0.151"
+quick-xml = { version = "0.41.0", features = ["serialize", "overlapped-lists"] }
 log = "0.4.28"
 chrono = { workspace = true }
 base64 = { workspace = true }
-rand = "0.10.0"
-rand08 = { version = "0.8", package = "rand" }
-uuid = "1.23.1"
-aes = "0.9.0"
+rand = "0.10.2"
+uuid = "1.24.0"
+aes = "0.9.2"
 keyring-core = { version = "1.0.0", features = ["sample"] }
-zerocopy = { version = "0.8.48", features = ["derive"] }
-bergshamra-c14n = "0.5.1"
-hmac = "0.12.0"
-bergshamra = "0.5.1"
-rsa = { version = "0.9.10", features = ["serde"] }
+zerocopy = { version = "0.8.55", features = ["derive"] }
+bergshamra-c14n = "0.8.0"
+hmac = "0.13.0"
+bergshamra = "0.8.0"
+rsa = { version = "0.9.10", features = ["serde", "sha2"] }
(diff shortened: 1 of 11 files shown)


crates/xodus/Cargo.toml
```diff
@@ -44,11 +45,8 @@ apple-native-keyring-store = { version = "1.0.0", features = ["keychain"]}
...
-bergshamra = "0.5.1"
-ed25519-dalek = "2.2.0"
+bergshamra = "0.8.0"
+ed25519-dalek = "3.0.0"
 
 [build-dependencies]
 prost-build = "0.14.4"
-
-[features]
-key-chain-file = []
```
imLinguin, line 54: Any reason the feature was removed?

crates/xodus/src/licensing/content.rs
imLinguin: Since MS-CV is added manually, the use of CorrelationVectorReqwestBuilder is no longer needed. I believe we could work arround this bug where that extension cant be applied, if we updated xal-rs libraries as well.

crates/xodus/src/licensing/utils.rs
```diff
@@ -26,36 +26,52 @@ pub fn parse_bcrypt_rsa_private(blob: &BCryptRsaBlock) -> rsa::errors::Result<Rs
...
+            let q_rsa = BigUint::from_bytes_be(&q_bytes);
+            RsaPrivateKey::from_components(n_rsa, e_rsa, d_rsa, vec![p_rsa, q_rsa])
         }
         RSAPRIVATE_MAGIC => {
             log::trace!("Got RSA Private");
             // No d in the blob — recompute it.
-            let one = BigUint::from(1u32);
-            let p1 = &p - &one;
-            let p2 = &q - &one;
+            let one = num_bigint_dig::BigUint::from(1u32);
```
imLinguin: nit: does this path have to be absolute? Sticks out from other uses of BigUint

imLinguin (changes requested):
few nitpicks

crates/xodus/Cargo.toml
BellezaEmporium: Might have been removed via cargo format, as I haven't touched that part myself.

crates/xodus/src/licensing/content.rs
BellezaEmporium: We could most certainly keep that line in, just as being commented for now while we see xal-rs.

crates/xodus/src/licensing/utils.rs
BellezaEmporium: Nope, unnecessary. Will import and name them to prevent any confusions between both rsa and num_bigint_dig

crates/xodus/Cargo.toml
imLinguin: That would explain reason behind 
```
warning: unexpected `cfg` condition value: `key-chain-file`
```

imLinguin (approved):

PR #115, opened 2026-08-07
feat: introduce xodus-game-proxy

imLinguin:


crates/xodus-game-proxy/src/main.rs
```diff
@@ -0,0 +1,41 @@
...
+
+#[cfg(target_os = "linux")]
+fn get_runtime_dir() -> String {
+    std::env::var("XDG_RUNTIME_DIR").expect("Runtime dir not set")
+}
+
+#[cfg(target_os = "macos")]
+fn get_runtime_dir() -> String {
+    "/tmp/".to_string()
+}
```
ChristopherHX, line 13: we shall put this in a new crate if this works, it is better to have those on one ipc package.

so we can never get a mismatch

imLinguin:
This didn't work as planned with wine. We'll revisit unixlib usage instead for xgameruntime

PR #118, opened 2026-08-11
fix: explain missing game license

### LINKED ISSUE #19: license requests for games that aren't owned need a better error
Instead of printing a error about some missing field "license" , better say something better formatted and clear up that you have to own a game you want a license for.

Koray-Ozt:
## Summary
- detect licensing API responses that omit or null out the `license` field
- return an actionable ownership/subscription error instead of the raw Serde `missing field license` message
- propagate the licensing error through the CLI rather than panicking
- add a regression test for the unowned-content response

## Test plan
- `cargo test -p xodus licensing::content::tests::missing_license_explains_ownership_requirement --lib`
- `cargo clippy -p xodus --lib --tests -- -D warnings`
- `cargo fmt --all --check`

Closes #19

### DIFF
--- a/crates/xodus-cli/src/license.rs
+++ b/crates/xodus-cli/src/license.rs
@@ -79,7 +79,7 @@ pub async fn get_license(
         market,
     )
     .await
-    .expect("failed to get license");
+    .map_err(|err| err.to_string())?;
 
     let game_splicense = SPLicense::parse_base64(&game_license.splicense_block)
         .expect("could not parse base64 game SPLicense");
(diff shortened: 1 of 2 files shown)


73chn1c:
Hey, just noticed this overlaps with #125 - we're both fixing #19, different approach though. Mine models the actual `satisfactionFailure` shape from the API (code + description), so it surfaces the real reason instead of a generic "you don't own this" message - might matter for stuff like regional restrictions or subscription mismatches too, not just missing ownership.

Not saying scrap yours, just flagging so whoever reviews doesn't get blindsided by two PRs touching the same lines. Happy to close mine if #125's approach isn't preferred, or they could get merged - up to whoever's reviewing.

Koray-Ozt:
Thanks @73chn1c! Modeling  via  in #125 is definitely cleaner, avoids double-parsing, and preserves the API's exact failure description. Closing this in favor of #125.

PR #119, opened 2026-08-11
fix(cli): pick default run executable deterministically

### LINKED ISSUE #106: Undeterministic default executable in xodus-cli run
* rusts hashmap order is random
* games like asphalt9 chose e.g. the crash utilities instead of the main .exe

---

We should actually support
- providing the absolute path to the executeable and backtrack it's MicrosoftGame.config and more

73chn1c:
## What

`xodus-cli run` without an explicit `--exe` picked its default launch target by taking "the first entry seen" while iterating the package's `HashMap<String, SegmentFile>`. Rust's default hasher randomizes iteration order per process, so the choice was non-deterministic across runs — matching the report of different executables being picked between runs.

## Fix

Break ties deterministically by picking the lexicographically smallest name among the `keep_encrypted` candidates. All `keep_encrypted` files are still mapped into `WINE_DLL_FILE_MAP` as before — only the *default launch target* selection changed.

(Earlier revision also filtered candidates to names ending in `.exe` — dropped per @ChristopherHX's comment that `keep_encrypted` implying `.exe` is already an invariant in this codebase, so the filter added nothing.)

## Testing

cargo build --workspace, cargo test --workspace, and cargo fmt --check --all all pass. No package test fixtures exist in this repo to exercise run.rs end-to-end, so this was verified by code inspection plus a standalone reproduction of the old vs. new selection logic against a synthetic file set (confirmed non-determinism before, determinism after).

Fixes #106

### DIFF
--- a/crates/xodus-cli/src/commands/run.rs
+++ b/crates/xodus-cli/src/commands/run.rs
@@ -214,16 +214,10 @@ pub async fn run(
 
     // HashMap iteration order is randomized per-process, so picking "the first
     // entry seen" as the default executable (when `--exe` isn't given) is
-    // undeterministic and can land on a non-.exe file. Pick deterministically
-    // instead, restricted to actual .exe candidates.
+    // undeterministic. Pick deterministically instead.
     let default_exe = exe
         .is_none()
-        .then(|| {
-            fds.iter()
-                .map(|(name, _)| *name)
-                .filter(|name| name.to_ascii_lowercase().ends_with(".exe"))
-                .min()
-        })
+        .then(|| fds.iter().map(|(name, _)| *name).min())
         .flatten();
 
     let mut nt_entry = None;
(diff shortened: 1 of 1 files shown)


ChristopherHX:
> Restrict default-selection candidates to files ending in `.exe` (case-insensitive), and break ties deterministically by picking the lexicographically smallest name. All `keep_encrypted` files are still mapped into `WINE_DLL_FILE_MAP` as before

Keep-Encryped = .exe is an unwritten law

73chn1c:
Good point, thanks — dropped the `.exe` filter, that invariant makes it unnecessary. Just the deterministic tie-break is left now.

73chn1c:
Withdrawing this — I never actually exercised this code path against a real downloaded package, only unit tests + build/fmt on unrelated code. That's not enough to claim it's fixed. Closing until I've verified it against real game data, will resubmit properly then.

PR #121, opened 2026-08-11
fix(hardware): bound the pkexec wait when probing SMBIOS on Linux

### LINKED ISSUE #120: load_raw_smbios() hangs forever if no polkit auth agent is running
## Bug

On Linux, `hardware.rs::load_raw_smbios()` shells out to:

```
pkexec cat /sys/firmware/dmi/entries/1-0/raw
```

with no timeout. If no polkit authentication agent is registered for the session (headless/SSH sessions, minimal window managers, containers, or just this specific desktop setup), `pkexec` blocks forever waiting for an authentication response that will never arrive. There's no error, no timeout — the calling binary (and anything awaiting `ensure_device_credentials`, i.e. every login/run path) just hangs indefinitely.

## Reproduction

Reproduced live on a plain Kubuntu 26.04 install with `libwebkit2gtk-4.1-dev`/`libgtk-3-dev` installed but no polkit desktop agent running in the shell that invokes the CLI:

```
cargo test --workspace
```

hangs indefinitely inside `api::live::test::test_get_xbox_live_dev_token` — `ps`/`pstree` on the stuck process tree shows a `pkexec` child sitting idle waiting on `do_wait`, never returning.

## Impact

Any user whose desktop session doesn't have a running polkit agent (or who runs xodus-cli/xodus-service outside a graphical session at all) gets a silent, indefinite hang the first time device credentials are provisioned, with zero indication of what's wrong.

## Suggested fix

Bound the wait on the `pkexec` child with a timeout, so a missing/non-responsive polkit agent degrades to a clean `io::Error` (which `probe_provision_components()` already handles via `.ok()`) instead of hanging forever. PR incoming.

73chn1c:
## What

`load_raw_smbios()` on Linux shells out to `pkexec cat /sys/firmware/dmi/entries/1-0/raw` and waits on it with `wait_with_output()`, unconditionally. If no polkit authentication agent is registered for the session, `pkexec` never returns — the wait is effectively infinite, with no error and no way to recover.

Reproduced live: on a plain desktop install with the CLI's build deps present but no polkit agent running in the invoking session, `cargo test --workspace` hangs forever inside `api::live::test::test_get_xbox_live_dev_token`, which calls `ensure_device_credentials` → `probe_provision_components` → `load_raw_smbios`. `pstree` on the stuck tree shows `pkexec` sitting idle on `do_wait`.

## Fix

Poll the child with `try_wait()` under a 10s bound instead of `wait_with_output()`. Stdout is read only after the child exits — safe here since the DMI type-1 raw entry is a few hundred bytes at most, well under a pipe buffer, so there's no concurrent-read deadlock risk to worry about. On timeout, the child is killed and a plain `io::Error(TimedOut)` is returned, which `probe_provision_components()` already tolerates via `.ok()` — callers see a `None` SMBIOS reading exactly like any other probe failure, just no longer after waiting forever.

## Testing

`cargo build --workspace`, `cargo fmt --check --all` pass. Before the fix, `cargo test -p xodus api::live::test::test_get_xbox_live_dev_token -- --include-ignored` hung indefinitely on this machine (no polkit agent in this session); after the fix it completes in the bounded time (SMBIOS probe times out at 10s, then the test proceeds to its real live-network step and fails there instead, for an unrelated reason — this environment can't reach `login.live.com`, which is expected and outside this PR's scope).

Fixes #120

### DIFF
--- a/crates/xodus/src/hardware.rs
+++ b/crates/xodus/src/hardware.rs
@@ -152,19 +152,46 @@ fn load_raw_smbios() -> io::Result<Vec<u8>> {
 
 #[cfg(target_os = "linux")]
 fn load_raw_smbios() -> io::Result<Vec<u8>> {
-    let cmd = Command::new("pkexec")
+    use std::time::{Duration, Instant};
+
+    let mut child = Command::new("pkexec")
         .args(["cat", "/sys/firmware/dmi/entries/1-0/raw"])
         .stdout(Stdio::piped())
         .spawn()?;
-    let output = cmd.wait_with_output()?;
 
-    if output.status.success() {
-        Ok(output.stdout)
-    } else {
-        Err(io::Error::new(
-            io::ErrorKind::PermissionDenied,
-            "unable to probe SMBIOS data",
-        ))
+    // pkexec blocks forever waiting on a polkit authentication agent. If none is
+    // running (headless/SSH/minimal WM sessions), the process never returns and
+    // hangs the caller indefinitely with no error. Bound the wait so that case
+    // degrades to an error instead. The DMI type-1 "raw" entry is a few hundred
+    // bytes at most, well under a pipe buffer, so it's safe to read only after
+    // the child has exited.
+    const TIMEOUT: Duration = Duration::from_secs(10);
+    let start = Instant::now();
+    loop {
+        if let Some(status) = child.try_wait()? {
+            let mut output = Vec::new();
+            if let Some(mut stdout) = child.stdout.take() {
+                use std::io::Read;
+                stdout.read_to_end(&mut output)?;
+            }
+            return if status.success() {
+                Ok(output)
+            } else {
+                Err(io::Error::new(
+                    io::ErrorKind::PermissionDenied,
+                    "unable to probe SMBIOS data",
+                ))
+            };
+        }
+        if start.elapsed() > TIMEOUT {
+            let _ = child.kill();
+            let _ = child.wait();
+            return Err(io::Error::new(
+                io::ErrorKind::TimedOut,
+                "pkexec did not respond within 10s (no polkit authentication agent running?)",
+            ));
+        }
+        std::thread::sleep(Duration::from_millis(100));
     }
 }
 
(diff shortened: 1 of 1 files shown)


imLinguin:
This solution is an double-edged sword. What if user is slow to type their password within that period?

73chn1c:
@imLinguin good point, thanks for flagging that — a fixed timeout is a bad tradeoff either way (too short cuts off a real person, too long is just a slower version of the same hang).

Reworked this to not use a timeout at all. Instead it checks whether a polkit authentication agent (`polkit-gnome-authentication-agent-1`, `polkit-kde-authentication-agent-1`, `lxqt-policykit-agent`, `lxpolkit`, `mate-polkit`, `xfce-polkit`, `ukui-polkit`) is actually running before calling `pkexec`:

- No agent running → `pkexec` can never show a prompt and would hang forever with no way out, so this fails immediately with a clear error instead of trying.
- An agent is running → this is a real, human-driven auth prompt, so `pkexec` is called with no artificial bound at all. It always returns on its own once you answer the prompt (accept, wrong password exhausting retries, or dismiss), so there's nothing to time out.

Pushed the updated diff, let me know if this looks right to you.

BellezaEmporium:
> @imLinguin good point, thanks for flagging that — a fixed timeout is a bad tradeoff either way (too short cuts off a real person, too long is just a slower version of the same hang).
> 
> Reworked this to not use a timeout at all. Instead it checks whether a polkit authentication agent (`polkit-gnome-authentication-agent-1`, `polkit-kde-authentication-agent-1`, `lxqt-policykit-agent`, `lxpolkit`, `mate-polkit`, `xfce-polkit`, `ukui-polkit`) is actually running before calling `pkexec`:
> 
> * No agent running → `pkexec` can never show a prompt and would hang forever with no way out, so this fails immediately with a clear error instead of trying.
> * An agent is running → this is a real, human-driven auth prompt, so `pkexec` is called with no artificial bound at all. It always returns on its own once you answer the prompt (accept, wrong password exhausting retries, or dismiss), so there's nothing to time out.
> 
> Pushed the updated diff, let me know if this looks right to you.

Checks fails on macOS

image

73chn1c:
Thanks for flagging - real bug, not a flake: `polkit_agent_running()` (and its `KNOWN_POLKIT_AGENTS` const) used `std::process::{Command, Stdio}`, but that import is `#[cfg(target_os = "linux")]`-gated while the function itself wasn't, even though its only caller already is. Pushed a fix gating the function/const the same way. Should be green now.

PR #123, opened 2026-08-11
fix(licensing): handle satisfactionFailure instead of panicking

### LINKED ISSUE #122: get_license_content panics on satisfactionFailure response instead of returning a clean error
## Bug

`licensing::content::get_license_content` deserializes the licensing endpoint response directly into `LicenseContentResponse { license: LicenseContent }`, assuming every response contains a `license` field. When the account has no entitlement for the requested content, the server instead returns a `satisfactionFailure` object with no `license` field at all, e.g.:

```json
{"satisfactionFailure":{"alternateContentIds":[...],"code":1,"data":[],"description":"Users do not possess any satisfying entitlements for the content id in question.","remediationAction":"PresentUpsell","remediationProductSkus":[{"productId":"...","skuId":"0010"}]}}
```

Deserializing that into `LicenseContentResponse` fails with `missing field \`license\``, and since the call site does `.expect("failed to get license")`, this surfaces as an opaque panic instead of a clear "not entitled" message:

```
thread 'main' panicked at crates/xodus/src/licensing/content.rs:53:79:
called `Result::unwrap()` on an `Err` value: Error("missing field `license`", line: 1, column: 1169)
```

## Reproduction

Reproduced live via `xodus-cli streaming <productId> <dest>` against a real, currently-authenticated Game Pass Ultimate account for a title the account isn't (yet?) entitled to. Same underlying call is used by `xodus-cli run`'s license fetch (`crates/xodus-cli/src/license.rs`).

## Suggested fix

Make `LicenseContentResponse` an untagged enum (mirroring the existing `PackageResponse::Found`/`NotFound` pattern in `models/packagespc.rs`) with a `SatisfactionFailure` variant, and have `get_license_content` return a proper error (the `description` field is a genuinely useful, human-readable message already provided by the server) instead of panicking. PR incoming.

73chn1c:
## What

`get_license_content` deserialized the licensing endpoint response directly into `LicenseContentResponse { license: LicenseContent }`. When the account has no entitlement for the requested content, the server returns a `satisfactionFailure` object instead (no `license` field), which failed deserialization and panicked at the `.expect("failed to get license")` call site in `xodus-cli/src/license.rs` with an opaque `missing field \`license\`` error.

## Fix

- `LicenseContentResponse` is now an untagged enum with `Success { license }` and `SatisfactionFailure { satisfaction_failure }` variants, mirroring the existing `PackageResponse::Found`/`NotFound` pattern in `models/packagespc.rs`.
- Added `LicenseContentError` (`thiserror`) with a `NotEntitled { description }` variant carrying the server's own human-readable description.
- `get_license_content`'s return type changed from `reqwest::Result<(LicenseContentResponse, License)>` to `Result<(LicenseContent, License), LicenseContentError>`.
- The one call site (`xodus-cli/src/license.rs::get_license`) already returns `Result<_, String>` and already discarded the first tuple element (`let (_response, ...)`), so this is a `.map_err(...)?` swap, not a signature change ripple.

## Testing

Reproduced live against a real, currently-authenticated account for a title it isn't entitled to:

Before:
```
thread 'main' panicked at crates/xodus/src/licensing/content.rs:53:79:
called `Result::unwrap()` on an `Err` value: Error("missing field `license`", line: 1, column: 1169)
```

After:
```
not entitled to this content: Users do not possess any satisfying entitlements for the content id in question.
```
(clean error, process exits instead of panicking)

`cargo build --workspace`, `cargo test --workspace`, `cargo fmt --check --all` all pass.

Fixes #122

### DIFF
--- a/crates/xodus-cli/src/license.rs
+++ b/crates/xodus-cli/src/license.rs
@@ -70,7 +70,7 @@ pub async fn get_license(
         return Err("Unsupported token".to_string());
     };
 
-    let (_response, game_license) = xodus::licensing::content::get_license_content(
+    let (_content, game_license) = xodus::licensing::content::get_license_content(
         client,
         ms_device_token,
         user_token,
@@ -79,7 +79,7 @@ pub async fn get_license(
         market,
     )
     .await
-    .expect("failed to get license");
+    .map_err(|err| err.to_string())?;
 
     let game_splicense = SPLicense::parse_base64(&game_license.splicense_block)
         .expect("could not parse base64 game SPLicense");
--- a/crates/xodus/src/licensing/content.rs
+++ b/crates/xodus/src/licensing/content.rs
@@ -7,8 +7,20 @@ use xal::cvlib::CorrelationVector;
 use crate::licensing::utils;
 use crate::models::devicecredential::License;
 use crate::models::licensing::{
-    DeviceContext, LicenseContentRequest, LicenseContentResponse, LicenseUserIdentity,
+    DeviceContext, LicenseContent, LicenseContentRequest, LicenseContentResponse,
+    LicenseUserIdentity,
 };
+
+#[derive(Debug, thiserror::Error)]
+pub enum LicenseContentError {
+    #[error("request error: {0}")]
+    Request(#[from] reqwest::Error),
+
+    /// The account has no entitlement for the requested content (not owned, or
+    /// not covered by the account's current subscription tier).
+    #[error("not entitled to this content: {description}")]
+    NotEntitled { description: String },
+}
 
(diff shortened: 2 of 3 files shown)


73chn1c:
Withdrawing this — I only verified the error branch (satisfactionFailure) live. The success branch's restructured field access (`content.keys[0].value` etc.) was never exercised against a real successful license response, which is the far more common real-world path. That's not sufficient testing for a fix like this. Closing until I can verify it end-to-end with an actual entitled title, will resubmit properly then. The underlying bug report (#122) stands — it's a real, reproduced panic.

Issue #130, opened 2026-08-11
streaming.rs's run_reader/run_cli_reader panic instead of erroring on any parse/IO failure


73chn1c:
## Bug

`commands/streaming.rs`'s `run_reader`/`run_cli_reader` (and their shared parsing logic) use `.expect(...)`/`.unwrap()` throughout - 21 call sites in this one file - for essentially every fallible step: opening the cache file, parsing the XVD container, parsing package files, parsing segment metadata, parsing NTFS metadata, etc. Both functions return `()`, not a `Result`, so there's currently no way for a failure partway through to surface as anything other than a panic.

## Reproduction

Trivial to hit with any corrupt/non-msixvc local file:

```
echo "not a real msixvc file" > /tmp/fake.msixvc
xodus-cli extract /tmp/fake.msixvc /tmp/dest --market US
```

```
thread 'main' panicked at crates/xodus-cli/src/commands/streaming.rs:253:61:
no err: Io(Custom { kind: UnexpectedEof, error: "early eof" })
```

Found while testing PR #131 (which fixes the `File::open` unwrap in the same function, for the missing-file-path case - that one was a single, isolated, safely-testable change). This issue is about everything downstream of that (the actual XVD/package/segment parsing), which is a much bigger surface.

## Why I didn't just fix it

Converting this properly means changing `run_reader`/`run_cli_reader`'s return type from `()` to something like `Result<(), Box<dyn std::error::Error>>` (or an `ExitCode`) and threading that through every one of the 21 call sites, plus their callers. That's a real, invasive refactor across a function that's already fairly long, and most of the failure paths (corrupt XVD headers, truncated segment metadata, mid-stream IO errors) aren't easily reproducible/testable without either crafting several different kinds of malformed msixvc fixtures or mocking the streaming reader - more than I could responsibly verify live in one sitting. Didn't want to submit a large, only-partially-tested change for something this central to the download path.

Flagging it clearly instead so it's tracked, with a concrete repro, rather than leaving it undiscovered.

imLinguin:
Im in a process of refactoring the downloading code. So current state should be treated as temporary

73chn1c:
Good to know, thanks - I'll steer clear of streaming.rs/download refactoring then and won't submit anything else touching that path while it's in flux. Leaving this issue open for tracking/context.

PR #136, opened 2026-08-12
fix(msixvc): propagate 4 unwraps as errors instead of panicking

73chn1c:
## What

These are distinct from the crate's other unwraps/`todo!()`s, which are deliberate stops for package shapes that haven't been reverse engineered yet (per `crates/msixvc/CLAUDE.md` - not touching those, on purpose). The four here are all "the format and failure condition are already fully understood, but the code panicked on it anyway instead of using the `Result` it already had":

- `XvdFile::parse`: a seek used `.expect("Unable to seek")` while every other fallible step in the same function already uses `?` - no reason for this one to be different, and a seek past EOF on a truncated/corrupted download is a real way to hit it.
- `parse_user_package_files`, `parse_segment_metadata`: both decode a fixed-size on-disk UTF-16 filename field with `String::from_utf16(...).unwrap()`. A corrupted/truncated download can produce an unpaired surrogate here, which is a decode error (`std::string::FromUtf16Error`), not an unknown format - the function already returns `Result<_, Box<dyn Error>>` and uses `?` for every other step. Extracted into two named helpers (`decode_nul_terminated_utf16`, `decode_utf16`) with real unit tests against an unpaired surrogate.
- `parse_ntfs_segment_metadata`: `part.bytes_start()`/`bytes_len()` (`gpt` crate) return `Err` specifically on integer overflow when multiplying a partition's LBA by the logical block size (confirmed by reading `gpt` 4.1.0's source) - a real possibility if the drive image's GPT partition table is corrupted, not a case the `gpt` crate leaves unhandled.

## Important scope note (added after a second review pass)

**This does not, by itself, remove the panic a user of `xodus-cli` would see.** Every call site of these functions in `run.rs`/`streaming.rs` immediately does `.expect("ok")`/`.expect("no err")` on the result - so a real failure here still panics, just one call frame higher, at the exact `.expect()` calls already tracked in #130 (which I know you're mid-refactor on). What this PR actually does: makes `msixvc`'s own functions correctly return/propagate `Result` instead of panicking internally, with real test coverage for the two UTF-16 cases - a prerequisite for #130's refactor to have something clean to propagate, once it reaches these call sites, not a user-visible fix on its own. Didn't want to overstate this.

## Testing

- `decode_nul_terminated_utf16`/`decode_utf16`: real unit tests (`cargo +stable test -p msixvc`, 11/11 pass) against an unpaired UTF-16 surrogate (0xD800 with no low surrogate), confirming clean `Err` instead of a panic.
- The seek and GPT-overflow fixes: verified by type-checking + reading the underlying error conditions (stdlib/`gpt` 4.1.0 source) rather than a live repro - didn't want to risk a hand-built binary fixture being subtly wrong and calling that "tested."
- `cargo +stable build/test --workspace`, `cargo +stable clippy --workspace` (matching this repo's actual CI invocation) and `cargo +stable fmt --check --all` all clean.

### DIFF
--- a/crates/msixvc/src/xvd.rs
+++ b/crates/msixvc/src/xvd.rs
@@ -313,12 +313,6 @@ fn decode_nul_terminated_utf16(units: &[u16]) -> Result<String, std::string::Fro
     String::from_utf16(&units[..end])
 }
 
-/// Decodes a UTF-16 field with no NUL terminator (as read directly off disk
-/// for a segment's file path in `parse_segment_metadata`) into a `String`.
-fn decode_utf16(units: &[u16]) -> Result<String, std::string::FromUtf16Error> {
-    String::from_utf16(units)
-}
-
 impl XvdFile {
     pub fn content_id(&self) -> uuid::Uuid {
         self.header.vduid
@@ -531,7 +525,7 @@ impl XvdFile {
                 ))
                 .await?;
                 file.read_exact(buf.as_mut_bytes()).await?;
-                let file_name: String = decode_utf16(buf.as_slice())?;
+                let file_name: String = String::from_utf16(buf.as_slice())?;
                 let page_length = if segment.filesize == 0 {
                     1
                 } else {
@@ -1078,17 +1072,5 @@ mod tests {
         // this in an on-disk filename field.
         let field = [0xD800u16, 0];
         assert!(decode_nul_terminated_utf16(&field).is_err());
-    }
-
-    #[test]
-    fn decode_utf16_errors_cleanly_on_an_unpaired_surrogate() {
-        let units = [0xD800u16];
-        assert!(decode_utf16(&units).is_err());
-    }
-
-    #[test]
-    fn decode_utf16_happy_path() {
-        let units: Vec<u16> = "hello".encode_utf16().collect();
-        assert_eq!(decode_utf16(&units).expect("should decode"), "hello");
     }
 }
--- a/crates/msixvc/src/xvd.rs
+++ b/crates/msixvc/src/xvd.rs
@@ -353,9 +353,7 @@ impl XvdFile {
 
         // TODO: Check if we have proper content type
         if xvd_header.xvc_data_length > 0 {
-            file.seek(std::io::SeekFrom::Start(xvc_info_offset))
-                .await
-                .expect("Unable to seek");
+            file.seek(std::io::SeekFrom::Start(xvc_info_offset)).await?;
             let xvc_info = XvcInfo::read(&mut file).await?;
 
             let region_count = xvc_info.region_count;
@@ -473,7 +471,7 @@ impl XvdFile {
                     .iter()
                     .position(|&c| c == 0)
                     .unwrap_or(fullname.len());
-                let pfull_name: String = String::from_utf16(&fullname[..end]).unwrap();
+                let pfull_name: String = String::from_utf16(&fullname[..end])?;
 
                 files.insert(
                     pfull_name,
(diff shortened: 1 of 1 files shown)


73chn1c:
Update: added real unit tests for the two UTF-16 decode fixes (extracted into named helpers, tested against an unpaired surrogate) - the PR description's note about not having live/behavioral tests for those two is now outdated, they're covered. The seek and GPT-overflow fixes are still verified by type-checking + reading the underlying error conditions only, as described.

73chn1c:
Important correction, caught on a second review pass: every call site of these functions in `run.rs`/`streaming.rs` immediately does `.expect("ok")`/`.expect("no err")` on the result. So this PR makes `msixvc`'s own functions correctly return `Result` (and adds real tests for the two UTF-16 cases), but it does **not** by itself remove the user-visible panic - it just moves where it would happen one call frame up, into the exact `.expect()` calls already tracked in #130. I'd overstated this in the original description. Value here is: `msixvc` itself is now correctly error-propagating and has test coverage for two real failure modes, which is a prerequisite for #130's refactor to actually do something once it reaches these call sites - not a fix for the symptom on its own. Wanted to flag this myself rather than have it come up in review.

73chn1c:
the invitation link has expired

73chn1c:
(link's dead anyway, so here it is) — yeah, I've been using Claude to help write these fixes. I'm still the one reading the code, deciding what's worth fixing, testing everything, and reviewing every diff before it goes up. None of this gets generated and shipped without me actually understanding it.

On the clean-room worry: none of tonight's PRs touch how the XVD/MSIXVC format itself gets figured out. They're error-handling cleanup (unwrap() -> ?, pulling a couple of decode helpers out with tests) on top of parsing logic that was already there. No disassembly, no Microsoft source, no debug symbols, no ReactOS - just the existing crate code and public Rust docs. If I ever touched actual format-discovery work I'd flag that myself for extra scrutiny.

Happy to keep talking about it here.

imLinguin:
@73chn1c Thanks for being clear about this. I agree use of LLM for identifying and addressing such localized code issues is valid and I dont see much reason not to pull this code.

crates/msixvc/src/xvd.rs
```diff
@@ -304,6 +304,21 @@ pub struct SegmentFile {
...
+fn decode_nul_terminated_utf16(units: &[u16]) -> Result<String, std::string::FromUtf16Error> {
+    let end = units.iter().position(|&c| c == 0).unwrap_or(units.len());
+    String::from_utf16(&units[..end])
+}
+
+/// Decodes a UTF-16 field with no NUL terminator (as read directly off disk
+/// for a segment's file path in `parse_segment_metadata`) into a `String`.
+fn decode_utf16(units: &[u16]) -> Result<String, std::string::FromUtf16Error> {
+    String::from_utf16(units)
+}
```
olivi-r: This function seems redundant

crates/msixvc/src/xvd.rs
```diff
@@ -1034,3 +1043,52 @@ impl XvdFile {
...
+    fn decode_utf16_errors_cleanly_on_an_unpaired_surrogate() {
+        let units = [0xD800u16];
+        assert!(decode_utf16(&units).is_err());
+    }
+
+    #[test]
+    fn decode_utf16_happy_path() {
+        let units: Vec<u16> = "hello".encode_utf16().collect();
+        assert_eq!(decode_utf16(&units).expect("should decode"), "hello");
+    }
```
olivi-r: We don't need tests for standard library behaviour

olivi-r (changes requested):

73chn1c:
Good catch on both, fixed - inlined the one-line decode_utf16 wrapper into its single call site and removed it along with its two tests, which were really just re-testing String::from_utf16's own behavior. Kept decode_nul_terminated_utf16 as-is, that one has actual logic (finding the NUL, falling back to the whole slice) worth its tests.

Issue #91, opened 2026-07-23
Research EAppx


imLinguin:
This is UWP focused format, but I believe we xodus can help with efforts to preserve those games
Initial points of integration with EAppx are:
- support for parsing EAppx 
   -  relevant references: https://pastebin.com/zH5tet0b https://github.com/dalion619/UnpEax
- support for decrypting individual files from archive

TSRBerry:
Look at https://github.com/tuxuser/eappx-rs for parsing + decryption.
The keys can be obtained with CIKExtractor as well

TSRBerry:
This will require a lot of WinRT components to be implemented.

imLinguin:
I believe at least being able to understand the format and unpack/mount underlying files would be enough to enable developers across community to test WinRT patches with more software.  
Thank you for referring `eappx-rs` - it looks like it could be straightforward to integrate with existing xodus codebase, at least for the download and extract commands.

TSRBerry:
I've used storelib_rs to download eappx games from the MS Store before, eappx-rs only implements decryption right now

PR #138, opened 2026-08-12
feat: add extract-eappx CLI command for local EAppx/EMSIX unpacking

asdfghjkljdhue7869:
Initial step toward #91 - adds a local-only `extract-eappx` command (mirrors the existing `extract` command for msixvc) that parses and extracts EAppx/EMSIX packages via the eappx-rs crate (tuxuser/eappx-rs, referenced in the issue discussion).

Scope is intentionally narrow: no store lookup or download, just local parse + decrypt + extract. This covers the "understand the format and unpack the underlying files" goal called out in the issue, so WinRT patches can be tested against more software. Download integration (storelib_rs) can build on this in a follow-up.

**Testing:** verified against eappx-rs's own bundled `testdata/TestApp_1.0.3.0_x64.emsix` fixture using a new `--key-test` flag (mirrors makeappx's equivalent flag) - all 13 files extracted successfully, including the encrypted TestApp.exe, TestApp.dll, and clrcompression.dll.

Happy to adjust naming/structure to match project conventions.


imLinguin:
We should be able to get keys from licenses, I'll reserve local file keys only for well known test keys that users dump themselves.

I didn't familiarize myself with that crate, but does eappx-rs even expose ways to fill keys manually?

crates/xodus-cli/src/commands/extract_eappx.rs
```diff
@@ -0,0 +1,98 @@
...
+        }
+    }
+
+    if key_test {
+        let test_keyid = KeyId::Guid((
+            Uuid::parse_str("ddafcf67-7b2c-086d-302b-8adac1bdd3a7").unwrap(),
+            Uuid::parse_str("7d53aeb8-5922-f062-b1d7-7e09f5a187a0").unwrap(),
+        ));
+        let test_key = hex::decode(
+            "9fe75f879e95a5d7f3715c30fce71067fc346efd680fa25e3c737d76acb72b9d",
```
Vidi0: Please, don't paste keys here. Keys should be obtained from MS's servers, not copied raw here. This is very wrong.

Vidi0 (changes requested):

crates/xodus-cli/src/commands/extract_eappx.rs
imLinguin: Even if this key is a test key and is usually embedded in a binary we cannot accept this change.
Users should dump keys themselves, as I pointed in my other comment

tuxuser:
> ... but does eappx-rs even expose ways to fill keys manually?

Yes it does. Here a small section from its README:

> Check out https://learn.microsoft.com/en-us/windows/win32/appxpkg/make-appx-package--makeappx-exe-#to-decrypt-a-package-with-a-key-file for keyfile format.

Reference: https://github.com/tuxuser/eappx-rs/blob/master/README.md

asdfghjkljdhue7869:
Subject: Re: extract-eappx CLI command

Yes — eappx-rs's KeyCollection exposes manual key filling directly. It has an add() method that takes a KeyId (the GUID pair from the package header) plus the raw key bytes, and a from_reader/from_str path for loading a serialized keyfile. That's exactly what the --key-test flag in this PR uses today to inject a single known key without touching the store.
Agreed on the direction — real games should resolve keys through the license flow rather than a local file. This PR's local-file/--key-test path was only ever meant to cover the case you're describing: well-known test keys a user has dumped themselves, for exercising the parser/extractor without a live license. Happy to help wire actual license-derived key retrieval as a follow-up once there's a clear place to hook into the existing token/license handling — let me know if you want that scoped as a separate PR or folded into this one.

Sent with [Proton Mail](https://proton.me/mail/home) secure email.

imLinguin:
Let's remove the test key from the code. And as initial unpacking support this is great.

asdfghjkljdhue7869:
Removed the hardcoded test key. Verified extraction still works via --key-file with a locally-generated keyfile (not committed) containing the same test key content - same 13 files extract cleanly, including the encrypted TestApp.exe/TestApp.dll/clrcompression.dll.

imLinguin (approved):
For initial testing it's enough

crates/xodus-cli/src/commands/extract_eappx.rs
```diff
@@ -0,0 +1,80 @@
...
+        eprintln!("Failed to load keys: {e}");
+        return ExitCode::FAILURE;
+    }
+
+    if !outdir.exists() {
+        if let Err(e) = std::fs::create_dir_all(&outdir) {
+            eprintln!("Failed to create output directory {:?}: {e}", outdir);
+            return ExitCode::FAILURE;
+        }
+    }
```
Vidi0, line 68: The clippy warnings should be easy to fix.

Vidi0 (approved):

PR #139, opened 2026-08-12
fix(cli): deterministic default executable selection for xodus-cli run (#106)

### LINKED ISSUE #106: Undeterministic default executable in xodus-cli run

Koray-Ozt:
Fixes #106

### Summary of Changes

- **Deterministic File Order:** Sorts  by relative file path prior to processing so Rust HashMap iteration non-determinism does not cause random binary selection.
- **Smart Executable Scoring:** Implemented  heuristic that filters  files and scores candidates:
  - Penalizes common helper/uninstaller/crash-reporter executables (e.g. ).
  - Prefers shallower directory depths (root/near root binaries).
  - Breaks ties alphabetically for 100% deterministic selection across runs.
- **Unit Tests:** Added unit tests verifying  logic.

### DIFF
--- a/crates/xodus-cli/src/commands/run.rs
+++ b/crates/xodus-cli/src/commands/run.rs
@@ -208,28 +208,38 @@ pub async fn run(
         fds.push((file.0, stdf.into_raw_fd()));
     }
 
+    // Sort fds deterministically by path
+    fds.sort_by(|a, b| a.0.cmp(b.0));
+
     let mut env_value = String::new();
     let nt_prefix = out_absolute.to_string_lossy().replace("/", "\\");
     let nt_prefix = nt_prefix.trim_end_matches('\\');
 
     let mut nt_entry = None;
 
-    for fd in fds {
+    let target_exe = match &exe {
+        Some(user_exe) => Some(user_exe.clone()),
+        None => {
+            let paths: Vec<&str> = fds.iter().map(|(p, _)| p.as_str()).collect();
+            select_default_exe(&paths)
+        }
+    };
+
+    for fd in &fds {
         if !env_value.is_empty() {
             env_value.push('|');
         }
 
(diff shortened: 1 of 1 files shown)


73chn1c:
Heads up that this overlaps with #124, which I opened for the same issue — flagging it only so the maintainers don't end up reviewing the same fix twice, not to stake a claim.

The two approaches differ: this one scores the `.exe` candidates heuristically, #124 reads `<ExecutableList>` out of the game's own `MicrosoftGame.Config` and uses the declared entry point.

One data point from testing a handful of titles locally that might be worth folding into the scoring here either way: Grounded's entry point is `Maine\Binaries\WinGDK\Maine-WinGDK-Shipping.exe`, so the "prefer shallower directory depth" rule would rank it below anything sitting at the tree root. The `update` / `tool` / `helper` keyword penalties have the same shape of risk for a game whose real executable happens to contain one of those substrings.

The `target_clean == nt_suffix` change is a good catch though, and independent of which selection strategy wins — `--exe` currently only matches the full backslash path from the manifest, so passing just a basename silently fails to match and you get "Could not find .exe". #124 doesn't fix that.

Koray-Ozt:
Thanks for the detailed comparison and the Grounded example. I agree that reading the package-declared entry point from `MicrosoftGame.config` is more reliable than scoring executable paths heuristically.

I’ll close this PR to avoid duplicating review effort with #124. The basename matching fix for an explicit `--exe` is independent and still useful; feel free to incorporate it into #124. Otherwise, I can submit it as a small, focused follow-up once #124 lands.

PR #140, opened 2026-08-12
feat(cli): add game playtime tracking (#92)

### LINKED ISSUE #92: PlayTimeTracking Support
Image
Tracking played time per game

Koray-Ozt:
Closes #92

## Summary

Adds persistent per-game playtime tracking to Xodus. Game sessions launched through `xodus-cli run` are measured and stored by content ID, and the new `xodus-cli playtime` command can display either a single title or a sorted summary.

## Changes

- Add `xodus::playtime::PlayTimeStore`, backed by a JSON file in the platform user-data directory:
  - Linux: `$XDG_DATA_HOME/xodus/playtime.json`, falling back to `~/.local/share/xodus/playtime.json`
  - macOS: `~/Library/Application Support/xodus/playtime.json`
- Measure playtime from successful child-process startup until process exit.
- Add `xodus-cli playtime [content_id]` without requiring keyring or device-credential initialization.
- Use atomic file replacement and an inter-process lock to prevent partial writes and lost concurrent updates.
- Preserve malformed stores instead of silently replacing them with empty data.
- Reject invalid data-directory configuration and playtime total overflow with explicit errors.
- Keep the feature Unix-only until equivalent atomic replacement behavior is implemented for Windows.

## Validation

- `cargo test --workspace`
  - `msixvc`: 6 passed
  - `xodus`: 14 passed, 1 ignored
  - `xodus-cli`: 2 passed
  - `xodus-service`: no tests, build passed
  - doc tests passed
- `cargo test -p xodus playtime::tests`: 9 passed
- `cargo fmt --all --check`
- `git diff --check`
- `cargo clippy -p xodus --all-targets -- -D warnings`
- `cargo clippy -p xodus-cli --all-targets -- -D warnings -A clippy::too_many_arguments`
  - The allowance is for two pre-existing streaming functions outside this change.
- Smoke-tested empty, populated, malformed, and missing-data-directory stores with the built CLI.


imLinguin:
I appreciate the effort here. However since we don't know how Xbox app handles gameplay itself. We should first explore how the official apps handles it and if the game doesn't update the similar statistic server side on its own.
It may be useful while offline, but we have not explored offline capabilities either

Koray-Ozt:
Thanks, that’s a fair concern. This implementation is intentionally local-only and does not update Xbox presence or profile statistics, but I agree that we should understand the official client and game behavior before introducing a second source of playtime data.

The traces added to #92 appear to describe the presence lifecycle, but they do not yet establish where accumulated playtime is calculated or persisted. I’m happy to close this PR for now so it does not remain in the review queue while that behavior is being investigated. If you think the local offline history is still useful as a separate feature, I can leave it open and adjust the scope and documentation accordingly.

PR #141, opened 2026-08-12
feat: track per-game playtime on session exit

### LINKED ISSUE #92: PlayTimeTracking Support

Empt-y:
## What

Adds a shared `xodus::playtime` module that persists accumulated playtime and last-played time per content id to a small JSON store under `XDG_DATA_HOME` (or the platform equivalent on macOS). `xodus-cli run` wraps the wine process wait with a session timer and records the elapsed duration once the game exits.

The store lives in the shared `xodus` crate rather than `xodus-cli` specifically so a future launcher UI or `xodus-service` can read/write the same data directly, rather than this being CLI-only.

## Testing

- Unit tests in `playtime.rs` covering: loading a missing store as empty, accumulating multiple sessions for the same and different content ids, and documenting (via a test) that sub-second sessions truncate to whole seconds rather than silently vanishing.
- `cargo test -p xodus playtime` - 3 passed.
- `cargo fmt --check` and `cargo clippy -p xodus --all-targets` - clean.

**Not verified**: I wasn't able to run `cargo build --workspace` / test `xodus-cli` directly - my environment is missing the `webkit2gtk-4.1` dev headers `wry` needs to build (the CLI's login webview dependency), unrelated to this change. I hand-checked `run.rs`'s usage against `xodus::playtime`'s public API (which does compile cleanly on its own), but this hasn't had a real end-to-end build. Happy to have someone with a full build confirm, or point me at what's needed to unblock it here.

Closes #92

### DIFF
--- a/crates/xodus-cli/src/commands/run.rs
+++ b/crates/xodus-cli/src/commands/run.rs
@@ -2,6 +2,7 @@ use std::collections::HashMap;
 use std::os::fd::{AsFd, IntoRawFd};
 use std::path::Path;
 use std::process::ExitCode;
+use std::time::Instant;
 
 use msixvc::models::xvd::PAGE_SIZE;
 use msixvc::xvd::{SegmentFile, XvdFile};
@@ -156,10 +157,12 @@ pub async fn run(
         lfiles.extend(sfiles);
     }
 
+    let content_id = xvd.content_id().to_string();
+
     let license = get_license(
         client,
         tokens,
-        xvd.content_id().to_string(),
+        content_id.clone(),
         market.unwrap_or("neutral".to_string()),
     )
     .await;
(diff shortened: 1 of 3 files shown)


Empt-y:
Closing this in favor of #140, which was opened first and covers the same issue (#92) with a more complete implementation:

- Ships an actual `xodus-cli playtime` command to view tracked time (this PR only persisted data with no way to inspect it)
- Guards concurrent writes with file locking (this PR has no locking - two simultaneous game sessions could race and lose updates)
- Writes atomically via temp file + rename (this PR uses a plain `fs::write`, so a crash mid-write can corrupt the store)
- Treats a corrupted store as an error instead of silently discarding it (this PR's `load` treats any parse failure as empty, so a corrupted file gets silently overwritten and loses all history on the next session)
- Wired into the existing "doesn't require device credentials" path other commands in this repo use for offline-only functionality

Didn't realize #140 was already up when I opened this - apologies for the duplicate effort. Happy to help review/test #140 if useful.

PR #147, opened 2026-08-13
fix(hardware): report unreadable SMBIOS fields as errors instead of panicking

### LINKED ISSUE #144: parse_smbios panics on unexpected DMI data instead of falling back to Component::error
`parse_smbios` indexes the raw DMI blob without any bounds or sanity checks, so device provisioning panics instead of falling back to the `Component::error(...)` path that's already there for exactly this case.

`probe_provision_components()` runs on the first launch of every command, and it's written to tolerate SMBIOS being unavailable — `load_raw_smbios().ok()` / `load_smbios_fields(...).ok()`, then `Component::error(4100/4101/4102)` when the result is `None`. That fallback works for `Err`, but `parse_smbios` reaches for fixed offsets and firmware-supplied indices directly, so anything unexpected in the blob aborts the process before the fallback can trigger.

Four separate spots, each reproduced with a unit test against `parse_smbios` on `main` (`b0609f2`):

| input | line | panic |
|---|---|---|
| empty blob | `hardware.rs:117` | `index out of bounds: the len is 0 but the index is 1` |
| SMBIOS 2.0 Type 1 (8 bytes — the UUID field only exists from 2.1 on) | `hardware.rs:121` | `range end index 24 out of range for slice of length 18` |
| `length` byte larger than the blob | `hardware.rs:123` | `range start index 127 out of range for slice of length 37` |
| string index larger than the number of strings published | `hardware.rs:142` | `index out of bounds: the len is 3 but the index is 7` |

The first one is reachable without any firmware weirdness at all: `pkexec cat /sys/firmware/dmi/entries/1-0/raw` can exit 0 with empty stdout, and `load_raw_smbios` only checks `status.success()`, so `Ok(vec![])` gets handed straight to `parse_smbios`.

Rows 2 and 4 are ordinary firmware variation rather than corruption — Type 1 was 8 bytes before SMBIOS 2.1, and the version/serial fields are 1-based string indices that firmware is free to set to anything (including `0`, meaning "not specified").

Happy path for reference — this one is fine and a fix shouldn't change it: a normal 2.4-layout Type 1 structure returns version, serial and the 16-byte UUID as expected.

Two things worth deciding as part of a fix:

1. Whether an unparseable blob should degrade to the existing `Component::error(...)` path (i.e. `parse_smbios` returns `io::Result`), which keeps provisioning working on odd hardware, or whether it should be a hard failure with a clear message.
2. Whether string index `0` should map to an empty string deliberately, rather than incidentally via the placeholder entry currently pushed at index 0.

Happy to put up a PR for whichever direction you prefer — I have the repro tests already written.

73chn1c:
Fixes #144, following the direction you gave there: each field gets read independently and falls back to `Component::error` on its own, and a string index of `0` is treated as a field error too.

### What changed

`parse_smbios` now returns a `SmbiosFields { version, serial, uuid }` with each field optional, and `probe_provision_components` pushes `Component::new` or `Component::error` per field. Previously all three were discarded together if anything went wrong — and, more to the point, most failures didn't reach that fallback at all because they panicked first.

The four inputs from the issue are now covered by tests, along with a truncated string set and the happy path:

- empty blob → no fields
- blob shorter than the SMBIOS 2.0 formatted section → no fields
- SMBIOS 2.0 Type 1 (8 bytes, no UUID field) → version + serial still read, `uuid` reported as an error
- `length` byte past the end of the blob → strings unreadable, but the UUID sits in the formatted section and is still returned
- string index past the string set, and index `0` → that field errors, the other one still comes through

### On the provisioning-flow concern

You asked whether the error fallbacks might upset the provisioning flow. I can answer half of it with evidence and want to be clear about the other half.

**Verified:** on this machine's real Type 1 structure, the new parser produces byte-identical `version`, `serial` and UUID values to the old one — I ran both implementations side by side over `/sys/firmware/dmi/entries/1-0/raw` and asserted equality. So on any machine that provisions successfully today, the request body is unchanged and there is no behavioural difference to test for.

**Not verified:** whether login.live.com still issues a device license when one or more of 4100/4101/4102 comes through as `Component::error`.

*(Correcting myself: the first version of this paragraph claimed testing it would consume one of the account's device slots. That was wrong — per #109 it's `DeviceAssoc.srf` that reserves one of the 10 slots, and xodus never calls it. `deviceaddcredential.srf` carries no user token at all, so provisioning a throwaway credential costs nothing on the account.)*

So it is testable, and I'd rather measure it than guess: post the same request with 4100/4101/4102 forced to `Component::error`, against a control run with the real values, and see whether `Success` and the license block still come back. I'll report the result here. Worth noting the existing code path already sends `Component::error` for 4101, 4160 and 4161 unconditionally, so the service clearly tolerates *some* error components — just not necessarily these.

If you'd prefer the conservative version in the meantime, I can make it fall back to sending all three as errors when any one of them fails (i.e. preserve the old all-or-nothing shape, minus the panics) — say the word and I'll push that instead.

### Testing

`cargo test --workspace` and `cargo fmt --check --all` pass locally on Linux.

One thing I could not check here: the `#[cfg(not(target_os = "linux"))]` branch of `load_smbios_fields` also changed shape to return `SmbiosFields`. `smbios-lib` is a Linux-excluded dependency, so I can't type-check that arm locally and I don't have a macOS cross-compile set up — the change is a mechanical transformation of the existing body (`ok_or_else(..)?.into_bytes()` → `.map(String::into_bytes)`), but the macOS CI job is the real check on it.

### DIFF
--- a/crates/xodus/src/hardware.rs
+++ b/crates/xodus/src/hardware.rs
@@ -13,6 +13,25 @@ use smbioslib::{SMBiosSystemInformation, SystemUuidData, table_load_from_device}
 use crate::clep;
 use crate::models::devicecredential::Component;
 
+/// The SMBIOS Type 1 fields sent during provisioning, each independently
+/// optional: firmware is free to omit any of them (a string index of 0 means
+/// "not specified"), and a structure predating SMBIOS 2.1 has no UUID field at
+/// all. A field that could not be read is reported as `Component::error`
+/// rather than aborting the whole probe.
+#[derive(Debug, Default, PartialEq, Eq)]
+pub(crate) struct SmbiosFields {
+    pub version: Option<Vec<u8>>,
+    pub serial: Option<Vec<u8>>,
+    pub uuid: Option<[u8; 16]>,
+}
+
+fn push_field(components: &mut Vec<Component>, id: u32, value: Option<impl AsRef<[u8]>>) {
+    match value {
+        Some(value) => components.push(Component::new(id, BASE64_STANDARD.encode(value))),
+        None => components.push(Component::error(id)),
+    }
+}
+
 pub fn probe_provision_components() -> Vec<Component> {
     let mut components = Vec::with_capacity(16);
     let drive_serial = BASE64_STANDARD.decode("AA==").unwrap();
@@ -20,7 +39,7 @@ pub fn probe_provision_components() -> Vec<Component> {
     let mut drive_buf = [0; 64];
 
     let smbios = load_raw_smbios().ok();
-    let parsed_smbios = load_smbios_fields(smbios.as_deref()).ok();
+    let parsed_smbios = load_smbios_fields(smbios.as_deref());
 
     drive_buf
         .iter_mut()
@@ -39,15 +58,9 @@ pub fn probe_provision_components() -> Vec<Component> {
     components.push(Component::new(8196, BASE64_STANDARD.encode(clepv2)));
     components.push(Component::new(8197, BASE64_STANDARD.encode(clepv4)));
 
-    if let Some((version, serial, uuid)) = parsed_smbios {
-        components.push(Component::new(4100, BASE64_STANDARD.encode(version)));
-        components.push(Component::new(4101, BASE64_STANDARD.encode(serial)));
-        components.push(Component::new(4102, BASE64_STANDARD.encode(uuid)));
-    } else {
-        components.push(Component::error(4100));
-        components.push(Component::error(4101));
-        components.push(Component::error(4102));
-    }
+    push_field(&mut components, 4100, parsed_smbios.version);
+    push_field(&mut components, 4101, parsed_smbios.serial);
+    push_field(&mut components, 4102, parsed_smbios.uuid);
 
     components.push(Component::new(4145, "AQAAAA==".to_string()));
     components.push(Component::error(4160));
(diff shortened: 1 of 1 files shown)


n7apollo:
Been tinkering with Claude Code in a local copy because I was curious about this project and I have a Linux gaming desktop running Bazzite, and I subscribe to Game Pass, and have been wanting to get Game Pass working. If this is at all helpful, here is one data point around this issue.

**Setup.** I build and run xodus inside a Fedora 44 distrobox on Bazzite. That container has no `pkexec`, so `load_raw_smbios()` fails immediately and `probe_provision_components()` takes the `else` branch. I confirmed that directly rather than inferring it — a scratch example calling `probe_provision_components()` on `main` prints:

```
 4113  ok
 4101  ERROR (-2147024894)
 8196  ok
 8197  ok
 4100  ERROR (-2147024894)
 4101  ERROR (-2147024894)
 4102  ERROR (-2147024894)
 4145  ok
 4160  ERROR (-2147024894)
 4161  ERROR (-2147024894)
```

**Result: provisioning and licensing both succeed in that state.** `xodus-cli login` completes and stores device + user tokens. `xodus-cli license <content-id> <out>` then returns a real CIK from `licensing.mp.microsoft.com` — a fresh HTTP/2 round trip, not a cached license. Tested against *A Game About Digging A Hole* (`9NGLST31DG26`, content id `6a14e5ac-f6f4-49f1-bde5-71f0eebff23f`).

So all three of 4100/4101/4102 arriving as `Component::error` is tolerated by the service.

**Caveats:**

- The device credential was minted by `login` in the same `pkexec`-less environment, so it was provisioned with the error components too — but what I instrumented is the component list the code produces, not the outgoing `deviceaddcredential.srf` body.
- One account, one machine: AMD RX 6600 XT (radeonsi, Mesa 26.2.1), Bazzite 44 / KWin 6.7.4 Wayland, Fedora 44 container.
- This is `main`, not your branch. It measures current behaviour when SMBIOS is unavailable, which is the same input your `else` path produces — it does not exercise your per-field split, where some fields succeed and others don't.

PR #149, opened 2026-08-13
fix(cli): match --exe by basename

Koray-Ozt:
## Summary
- match an explicit `--exe` against either the full manifest path or its basename
- support both backslash and forward-slash path separators
- add a regression test for nested WinGDK executable paths

This is independent of the default executable selection in #124. Without it, `--exe Game.exe` does not match a manifest path such as `\Binaries\WinGDK\Game.exe` and `xodus-cli run` exits with `Could not find .exe`.

## Testing
- `cargo test --workspace`
- `cargo fmt --all --check`
- `cargo clippy --workspace` (passes with two existing `too_many_arguments` warnings in `streaming.rs`)

### DIFF
--- a/crates/xodus-cli/src/commands/run.rs
+++ b/crates/xodus-cli/src/commands/run.rs
@@ -33,6 +33,14 @@ fn make_temp_file(folder: &str) -> std::io::Result<std::fs::File> {
     }
 }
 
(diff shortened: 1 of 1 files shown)


Koray-Ozt:
Withdrawing this PR.

PR #152, opened 2026-08-17
fix(licensing): guard SPLicense id_len underflow and decode UTF-16 lossily

### LINKED ISSUE #145: SPLicense parser: unchecked id_len - 16 underflows (allocator abort in release), plus from_utf16 unwrap
`SPLicense::merge_tlv` takes two length fields straight out of the blob and uses them without checking, so a malformed or truncated SPLicense aborts the process instead of returning `SPLicenseDecodeError`. Both spots are in the same function; the surrounding code already has a proper error enum, so this looks like an oversight rather than a deliberate choice.

Reproduced against `main` (`b0609f2`) with unit tests driving `SPLicense::decode` on synthetic blobs.

### 1. `id_len - 16` underflows (`splicense.rs:275`)

```rust
let id_len = read_u16(&mut reader)? as usize;
// key_len is always 40
let _key_len = read_u16(&mut reader)? as usize;

let key_id = read_uuid(&mut reader)?;
let _unknown = read_vec(&mut reader, id_len - 16)?;
```

`id_len` is a `u16` read from the `PackedContentKeys` block. Nothing constrains it to `>= 16`, and the subtraction is unchecked:

- **debug** — `splicense.rs:275:58`, `attempt to subtract with overflow`
- **release** — wraps to ~1.8e19, `read_vec` does `vec![0u8; len]`, and it aborts in `library/alloc/src/raw_vec/mod.rs:28:5` with `capacity overflow`

The release path is the worse of the two: it's an abort from inside the allocator, so a caller can't turn it into a clean error no matter how it handles the `Result`.

The `offset` accounting in the same loop is fine, for what it's worth — `offset += 4 + id_len + 40` matches the `2 + 2 + 16 + (id_len - 16) + 40` actually consumed, so only the subtraction needs guarding.

### 2. `String::from_utf16(...).unwrap()` (`splicense.rs:260`)

The `PackageFullName` block is decoded with `.unwrap()`, so an unpaired surrogate panics with `called Result::unwrap() on an Err value: FromUtf16Error(())`. Same class as the UTF-16 unwraps in #136, just in a different file — `from_utf16_lossy`, or propagating through `SPLicenseDecodeError`, would both work.

### Checked and *not* a problem

I also suspected `read_vec(&mut reader, size - 4)` at `:289` (`SignatureBlock`) of the same underflow, but it isn't reachable: with `size < 4` the preceding `read_array`/`read_u16` hit EOF on the `take(size)`-limited reader first and return a clean `Err`. Tested, passes — mentioning it so nobody re-checks that line.

### Scope

To be clear about what I have and haven't shown: I have not observed a real license blob from the service with `id_len < 16`, so this is hardening the parser against malformed input rather than a failure anyone is hitting today. It matters because `SPLicense::parse_base64` is fed directly from server responses in `tokens/device.rs` and `xodus-cli/src/license.rs`, and because the release-mode failure is an allocator abort rather than a catchable error.

I have the repro tests written and can open a PR if you'd like — though this is your parser and you may prefer to decide how strict it should be (reject the block, skip it, or clamp) before anyone writes the fix.

snowyukitty:
### Summary
This pull request hardens SPLicense parsing in \crates/xodus/src/licensing/splicense.rs\ against arithmetic underflow and malformed UTF-16 panics (#145):

1. **Underflow Guard**: Added a guard check \id_len < 16\ in \BlockId::PackedContentKeys\ before evaluating \id_len - 16\, returning an \io::ErrorKind::InvalidData\ error cleanly instead of causing an arithmetic underflow / allocator abort.
2. **Safe UTF-16 Decoding**: In \BlockId::PackageFullName\, replaced \String::from_utf16(...).unwrap()\ with \String::from_utf16_lossy(...)\ to safely decode invalid surrogate sequences without panicking.
3. **Unit Tests**: Added regression tests covering malformed/underflow blobs, valid content key blocks, malformed surrogate UTF-16 decoding, and valid package full name strings.

Fixes #145

### DIFF
--- a/crates/xodus/src/licensing/splicense.rs
+++ b/crates/xodus/src/licensing/splicense.rs
@@ -200,6 +200,15 @@ pub enum SPLicenseDecodeError {
 
     #[error("expected to read {expected} bytes but only {read} were read")]
     PayloadLengthMismatch { expected: usize, read: usize },
+
+    #[error("PackedContentKey id_len {id_len} is less than 16")]
+    InvalidPackedContentKeyIdLength { id_len: usize },
+
+    #[error("invalid UTF-16 package name byte length {len}")]
+    InvalidPackageNameByteLength { len: usize },
+
+    #[error("invalid UTF-16 package name: {0}")]
+    InvalidPackageNameUtf16(#[from] std::string::FromUtf16Error),
 }
 
 #[derive(Debug, Error)]
@@ -253,11 +262,16 @@ impl SPLicense {
             }
             Ok(BlockId::PackageFullName) => {
                 let data = read_vec(&mut reader, size)?;
+                if data.len() % 2 != 0 {
+                    return Err(SPLicenseDecodeError::InvalidPackageNameByteLength {
+                        len: data.len(),
+                    });
+                }
                 let utf16: Vec<u16> = data
                     .chunks_exact(2)
                     .map(|chunk| u16::from_le_bytes([chunk[0], chunk[1]]))
                     .collect();
-                let mut s = String::from_utf16_lossy(&utf16);
+                let mut s = String::from_utf16(&utf16)?;
                 if s.ends_with('\0') {
                     s.pop();
                 }
@@ -272,10 +286,9 @@ impl SPLicense {
                     let _key_len = read_u16(&mut reader)? as usize;
 
                     if id_len < 16 {
-                        return Err(io::Error::new(
-                            io::ErrorKind::InvalidData,
-                            format!("PackedContentKey id_len {id_len} is less than 16"),
-                        ).into());
+                        return Err(SPLicenseDecodeError::InvalidPackedContentKeyIdLength {
+                            id_len,
+                        });
                     }
 
                     let key_id = read_uuid(&mut reader)?;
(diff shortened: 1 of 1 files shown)


imLinguin (approved):
Thanks, the changes look good. Please fix the formatting before we can merge

crates/xodus/src/licensing/splicense.rs
```diff
@@ -271,6 +271,13 @@ impl SPLicense {
                     // key_len is always 40
                     let _key_len = read_u16(&mut reader)? as usize;
 
+                    if id_len < 16 {
+                        return Err(io::Error::new(
```
Vidi0: This should have its own error variant in `SPLicenseDecodeError`, not a general io Error.

crates/xodus/src/licensing/splicense.rs
```diff
@@ -458,3 +465,97 @@ impl Deref for ContentKey {
         &self.0
     }
 }
+
+#[cfg(test)]
+mod tests {
```
Vidi0, line 483: I don't that many tests are really needed...

crates/xodus/src/licensing/splicense.rs
```diff
@@ -257,7 +257,7 @@ impl SPLicense {
                     .chunks_exact(2)
                     .map(|chunk| u16::from_le_bytes([chunk[0], chunk[1]]))
                     .collect();
-                let mut s = String::from_utf16(&utf16).unwrap();
+                let mut s = String::from_utf16_lossy(&utf16);
```
Vidi0: If we're going to be strict, it's better to return a custom error variant of `SPLicenseDecodeError` on malformed input.

Vidi0 (changes requested):

snowyukitty:
Thanks @Vidi0 and @imLinguin for the thorough review!

Updated in ad5aa9:
1. Added dedicated SPLicenseDecodeError variants:
   - InvalidPackedContentKeyIdLength { id_len: usize } for id_len < 16.
   - InvalidPackageNameByteLength { len: usize } for odd package-name byte lengths.
   - InvalidPackageNameUtf16(#[from] std::string::FromUtf16Error) replacing lossy conversion to strictly reject malformed UTF-16 surrogate sequences.
2. Re-ran cargo fmt --all across the workspace.
3. Focused tests on these error paths alongside the positive UTF-16 package-name decode test; all unit tests pass cleanly.

Vidi0 (approved):

imLinguin:
Thank you!

PR #154, opened 2026-08-19
Issue tokens for the service a title is actually calling

Denkiyaa:
Titles that talk to their publisher's own back end could not sign in. That back end
checks two things we were not providing: that the token was issued for the relying
party Xbox registered for the service, and that it carries the title's claim.

We handed out a generic `http://xboxlive.com` token minted from the user alone, so:

- No Man's Sky's auth server answered **400** to every request
- PlayFab answered **"The Xbox token provided is not valid for title 306180"**

## Where those relying parties live

Only in the **title-scoped** endpoint document at `titles/current/endpoints`. Worth
recording, because none of it is obvious:

- The platform-wide `titles/default/endpoints?type=1` document (88 entries) contains no
  third-party host at all
- `titles/current/endpoints` needs a token carrying the title's claim; a plain user
  token gets **403**
- That route also **rejects** the `type` query the default document requires - with it
  you get 414

For No Man's Sky it answers:

```json
{"Host":"*.nomanssky.com","RelyingParty":"https://nomanssky.com/","TokenType":"JWT"}
```

`get_endpoint` already knew how to map a URL onto such a document, but nothing called it.

## What this changes

A caller may now name the service URL where a relying party goes. The title document
decides, falling back to the platform-wide one and then to the string unchanged, so
callers that already hold a relying party are unaffected.

Tokens are minted through a SISU flow for the title. That flow runs **once per session**:
doing it per relying party made sign-in slow enough that No Man's Sky gave up and retried
before the first token landed, and sent PlayFab a null token.

`TitleMgtResponse::signature_policies` became optional because title-scoped documents
list endpoints only.

## Measured

| | before | after |
|---|---|---|
| `merged-nms-auth.nomanssky.com` | 400 | **200**, JWT returned |
| PlayFab `LoginWithXbox` | `InvalidXboxLiveToken` | **SessionTicket** |
| RV There Yet? | "could not sign in" | **signs in, reaches gameplay** |

Tested against Minecraft, No Man's Sky and RV There Yet? under WineGDK. The matching
xgameruntime side (the request URL now travels to the service, and the title id goes out
with it) is in Weather-OS/WineGDK.

No Man's Sky still stops at its own "press to begin" screen for an unrelated reason: it
posts to `qa-nms-auth` rather than the production host, chosen somewhere in its own data.
Its identity chain is complete.

### DIFF
--- a/crates/xodus-cli/src/commands/download.rs
+++ b/crates/xodus-cli/src/commands/download.rs
@@ -1,12 +1,18 @@
+use std::io::SeekFrom;
 use std::process::ExitCode;
+use std::time::Duration;
 
 use futures_util::StreamExt;
 use indicatif::{ProgressBar, ProgressStyle};
 use inquire::MultiSelect;
 use inquire::validator::Validation;
-use tokio::io::AsyncWriteExt;
+use reqwest::header::RANGE;
+use tokio::io::{AsyncSeekExt, AsyncWriteExt};
 use xodus::models::packagespc::PackageFile;
 use xodus::tokens::TokenManager;
+
+/// Consecutive attempts that move no bytes before a download is called off.
+const MAX_DOWNLOAD_ATTEMPTS: u32 = 10;
 
 use crate::package::{get_content_id, get_packages};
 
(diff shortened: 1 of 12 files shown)


imLinguin:
We already have unmerged changes in https://github.com/xodus-gaming/xgameruntime aligned with our initial idea of xodus providing live tokens to xgameruntime, that are then handled in runtime session itself - together with request signatures.

Additionally, while its not much of Xodus problem, as a lot of this was already researched and provided as scattered utilities, on the other hand WineGDK will be unable to pull your changes due to them being assisted by an LLM - which is in a direct violation of Wine clean room guidelines - https://gitlab.winehq.org/wine/wine/-/wikis/Clean-Room-Guidelines

Denkiyaa:
Thanks for the clear explanation. I opened these without knowing.

PR #155, opened 2026-08-20
Update README about supported Rust version

Vidi0:
To be able to use the newest stable Rust features, like the new range API (https://doc.rust-lang.org/std/range/index.html), the latest stable Rust compiler is required. I think it's useful to be able to use those latest features, so this change makes the compiler requirement explicit in the README.

I think it's not problematic to ask for the latest *stable* Rust compiler, because it can be easily downloaded via rustup. Also, the API added in a stable Rust compiler is guaranteed to keep existing in the next releases (it's not like nightly, where features are added and removed freely).

### DIFF
--- a/README.md
+++ b/README.md
@@ -59,7 +59,7 @@ The project structure is as follows.
 
 ### Prerequisites
 
-- Rust version supporting `edition = "2024"`
+- Latest stable Rust version
 - Right now CLI relies on wry and tao to show a login page. Consult https://docs.rs/wry/latest/wry/#platform-considerations
 - xodus-service relies on `protoc` to compile `proto/` definitions make sure to install it for your platform
 
--- a/README.md
+++ b/README.md
@@ -59,7 +59,7 @@ The project structure is as follows.
 
 ### Prerequisites
 
-- Latest stable Rust version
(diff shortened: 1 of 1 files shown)


imLinguin (approved):

BellezaEmporium (commented):
Still, a minimal edition should be provided. Not everyone will constantly update their Rust version.

(so, technically, 1.96.0)

imLinguin:
That's a valid point. std::Range is currently the newest feature we require, and coincidentally was stabilized in recent 1.96.0

1.98.0 is the newest, but there were no relevant features we require.

Vidi0:
Well, it would be useful to use the new [`String::from_utf16le`](https://doc.rust-lang.org/std/string/struct.String.html#method.from_utf16le) stabilized in Rust 1.98.

imLinguin:
Let's be specific as per to the version nevertheless

Vidi0:
Updated so the requirement is Rust 1.98 or later.

imLinguin (approved):

BellezaEmporium:
Nice job mates. 🤝

Issue #32, opened 2026-06-13
Extract needs resumption


Lariaa:
If you are stuck in a really large file on a really large game (or build a debug build by accident) it is always good to be able to pickup where you left off.

ChristopherHX:
https://github.com/xodus-gaming/xodus/pull/40 is an experiment that will identify where something is missing / incomplete or corrupted.

PR #156, opened 2026-08-20
Resumable downloads, reclaimed install space, and the Xbox Live token chain

MYCl0238:
The CLI and service side of running Game Pass titles end to end. Companion to
xodus-gaming/xgameruntime#10 and xodus-gaming/wine#3.

### Downloads resume, and installs stop costing twice the disk

- **An interrupted download continues from its cache** rather than starting
  over (#32). These packages run to tens of gigabytes, and losing all of it to
  a closed lid or an unplugged drive is the difference between an inconvenience
  and an unusable installer.
- **The installed package hands its payload back.** An install was ~2x the
  download because the whole MSIXVC was kept beside the extracted game, though
  only ~0.2% of it is ever read again. Punching out the payload past the prefix
  a launch needs took a library here from 135 GB to 69 GB, with every title
  still launching.
- Exit codes mean something. Five failure paths printed a message and returned
  normally while `run()` returned success unconditionally, so a licence refusal
  exited 0 and an install that produced nothing looked like it had worked.

### The Xbox Live token chain

- The relying party is resolved from Xbox's endpoint table rather than guessed
  from the URL, and the request a title is about to send is signed — including
  its body, which a presence POST needs and which nothing had sent before.
- MSA faults report their actual fault code instead of surfacing as a
  deserialisation failure, and a title ticket MSA has already refused is not
  asked for again on a loop.
- **This overlaps #108**, which mints title-claimed XSTS tokens over IPC and
  touches the same two files. Ours grew from the title side, against real
  titles' `XUserGetTokenAndSignature` calls; if #108 lands first, the signing
  work here rebases onto it cleanly and I am glad to do that.

### Launching

- **The executable is chosen deterministically** (#106). Hades II ships
  `F10.exe` beside `Hades2.exe` and the launcher started the wrong one, so the
  package manifest's `<Application Executable=...>` decides, with a ranking
  fallback for packages that do not declare one. **This overlaps #124**, which
  reads `MicrosoftGame.config` for the same purpose and adds `quick-xml` and
  `serde` to do it; this version needs no new dependencies but theirs reads the
  GDK's own manifest, and I have no attachment to mine winning.
- Arguments after `--` reach the title.

### Verified

Fifteen catalogue titles install and launch, with sign-in, achievements
read and write, real-time activity and presence working against a real Xbox
account.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -370,16 +370,16 @@ dependencies = [
  "const-oid 0.10.2",
  "der 0.7.10",
  "dsa",
- "ecdsa",
+ "ecdsa 0.16.9",
  "ed25519-dalek 2.2.0",
  "kryptering",
  "ml-dsa",
  "num-bigint-dig 0.8.6",
  "num-traits",
- "p256",
+ "p256 0.13.2",
  "p384",
  "p521",
- "pem-rfc7468",
+ "pem-rfc7468 0.7.0",
  "pkcs1",
  "pkcs8 0.10.2",
  "pkcs8 0.11.0",
@@ -941,6 +941,7 @@ checksum = "1a52aa3fcda4e6302a9f48734f234d35d4721b96f8fe07d073f07ce9df4f0271"
 dependencies = [
  "cpubits",
  "ctutils",
+ "getrandom 0.4.3",
  "hybrid-array",
  "num-traits",
  "rand_core 0.10.1",
@@ -1190,7 +1191,7 @@ dependencies = [
  "const-oid 0.9.6",
  "der_derive",
  "flagset",
- "pem-rfc7468",
+ "pem-rfc7468 0.7.0",
  "zeroize",
 ]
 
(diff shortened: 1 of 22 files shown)


imLinguin:
Extract and download commands are debug only, the current CLI isnt meant to be used yet so it shouldnt be treated as a stable interface. We are using it to iterate and test given parts e2e on various games.
The assumption that msixvc will be downloaded as a whole is wrong, we are going to keep only `RESIDENT` sectors as I'm refactoring the download pipeline for performance and scalability with other formats we have in mind like msixvc2, and UWP's eappx and msix we are planning on researching in the future. 

The pull request touches a lot of unrelated parts at once, this makes it harder to digest for us reviewers, but heavy use of AI and completely pointless comments in the code don't look that plausible for us to consider. 
Even if some bits are good quality I wouldnt accept all the changes as they are now.

You admit that some of your changes overlap with other PRs. How does it make it easier for us? Why are the changes even there if other PRs already cover them?

PR #157, opened 2026-08-22
Add new parser for fixed-length binary structures

Vidi0:
This PR adds a new parser, `BytesReader`, which is used in order to parse fixed-length binary structures from bytes into Rust structs.

The main advantage of the parser is that it tracks the remaining bytes at compile time (using typenum's `Unsigned`), so it is impossible to do out-of-bounds reads (and also it is impossible to leave bytes without being read, every byte must be read or discarded due to how the API was made). Also, every struct that can be parsed and implements `BinaryParse` or `BinaryTryParse` can be read from another `BytesReader` via `BytesReader::read` or `BytesReader::try_read`, respectively. One last thing I really like about the API is that arrays (or `GenericArray`s) of types implementing `BinaryParse` also implement `BinaryParse` directly (also works for the fallible `BinaryTryParse` version).

These types can be used to parse any kind of fixed-length binary structure. 

Also, it works over references, so for example when parsing an array of the same fixed-length struct, the buffer can be reused. `BytesReader` isn't as helpful for parsing variable-length structs, but it can be used in order to parse fixed-length structs from inside variable-length ones.

Examples of the usage of this parser can be find at `crates/msixvc/src/models`. This replaces the old way of parsing, which consisted of first transmuting bytes into a raw struct and then parsing into a better representation, now everything's done in one go.


imLinguin (approved):
I really like this change, especially how we are not removing duplication of structs `raw` and `parsed` pairs. 
Not 100% sure if worth pursuing, but in theory we could try adding a derive macro to some of those traits in the future if we wanted to cut down on boilerplate further.

PR #158, opened 2026-08-24
chore: update dependencies

Vidi0:
Also, I added `thiserror` and `uuid` as workspace dependencies, as they're used in almost every crate in the workspace.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -36,16 +36,16 @@ dependencies = [
 
 [[package]]
 name = "aes-gcm"
-version = "0.11.0"
+version = "0.11.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "fdf011db2e21ce0d575593d749db5554b47fed37aff429e4dc50bc91ac93a028"
+checksum = "7f2b8006a0c83f52b62ba44a97b58bf76fe2f70a329e588f67f89691d93d498f"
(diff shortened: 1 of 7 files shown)


imLinguin:
Did you verify all login.live services still work as expected? quick-xml was bumped so it's wise to double-check the data is in the format we expect

Vidi0:
None of the breaking changes in the [release notes](https://github.com/tafia/quick-xml/releases/tag/v0.42.0) affect the API that we use, so I don't think there should be any problem.

ChristopherHX:
IMHO we should run more than lint in ci

device auth flow does not need credentials

Vidi0:
> IMHO we should run more than lint in ci
> 
> device auth flow does not need credentials

We should also run `cargo test`, but for that to be useful we would need to write the tests. We can still add it to CI anyways.

ChristopherHX:
I would try today in about 4 hours
- to enable CI tests that check xml device auth + xbl device token
- test detection old device key in keyring

Then rebase this and merge if all good for CI as well.

Any comments welcome

ChristopherHX (approved):
xbox device auth and getting a token is completely fine with that change, I have no objections for this change

while I also should bump dependencies in xal-rs and then update it here. So both use same version of reqwest is used on both sides. To be discussed later.

PR #161, opened 2026-08-24
chore: fix lint

Vidi0:
Now that the required Rust version has been bumped, use the new `String::from_utf16le` function. Also, simplify another thing that was triggering clippy.

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -29,6 +29,8 @@ jobs:
       - name: Install protobuf
         if: ${{ runner.os == 'macOS' }}
         run: brew install protobuf
+      - name: Update Rust toolchain
+        run: rustup update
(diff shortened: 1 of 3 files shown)


Vidi0:
It seems like the GitHub runner hasn't been updated to Rust 1.98 yet.

ChristopherHX:
should I amend this pr to install the correct toolchain?

Vidi0:
@ChristopherHX Sure!

Vidi0:
Now the Rust toolchain is updated on CI.

Vidi0:
Should parsing from a UTF-16 string retain any null character at the end? The old implementation removed them, but it's not required. Leaving them would simplify the parsing logic. Now it retains those null bytes.

imLinguin (approved):

PR #163, opened 2026-08-27
feat: get_license_token api for XStore

ChristopherHX:


imLinguin (approved):
Looks great, clippy had some warnings, should be simple to resolve, then we can merge

PR #165, opened 2026-08-29
Improve MSIXVC layout handling [1/2]

Vidi0:
This is the first step towards a better handling of the layout of MSIXVC binaries.

This PR adds a new `layout` module, which contains the `Bytes` and `Pages` structs. Both structs are newtype wrappers over an u64 and have associated methods for converting between them (`Bytes` to `Pages` have two conversions, one for flooring and other for ceiling, while `Pages` to `Bytes` has only one). The `PAGE_SIZE` and similar constants have been moved to the `layout` module. The idea is to have a type-level distinction between a number of bytes and a number of pages, and also allows removing some functions converting bytes to pages and vice versa from the `math` module.

The next step would be precomputing the layout of a MSIXVC binary based on the header into a `XvdLayout` struct, and storing the layout alongside the header into `XvdFile`. This layout struct would contain the page range of each section of the file (MDU, drive, xvc info, etc.). I'm still unsure about how to store the layout, so I'll send the changes in a later patch.

### DIFF
--- a/crates/msixvc/src/crypt.rs
+++ b/crates/msixvc/src/crypt.rs
@@ -1,4 +1,5 @@
-use crate::models::xvd::{PAGE_SIZE, XvcRegionId};
+use crate::layout::PAGE_SIZE;
+use crate::models::xvd::XvcRegionId;
 
 use std::iter;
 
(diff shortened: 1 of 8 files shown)


imLinguin (approved):
This is awesome, will definitely let us avoid potential calculation issues as long as we enforce the type use where possible

PR #168, opened 2026-09-01
Improve MSIXVC layout handling [2/2]

Vidi0:
I wasn't able to remove the remaining stuff in math.rs (yet), but this moves all the other computations to `models::xvd::layout`. The layout of the MSIXVC package is computed once at the start. The layout includes the start page and length in bytes of each section of the package, plus the layout of the Hash Tree.

### DIFF
--- a/crates/msixvc-common/src/parse/byteorder.rs
+++ b/crates/msixvc-common/src/parse/byteorder.rs
@@ -10,10 +10,10 @@ use typenum::{U2, U4, U8, U16};
 
 use std::marker::PhantomData;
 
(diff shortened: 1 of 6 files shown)


Vidi0:
(I didn't bother updating the code to use the new `HashTreeLayout` struct because the Hash Tree handling needs to be reworked anyways, this way the review is simpler)

imLinguin (approved):
Looking good While updating the use of HashTreeLevels makes sense to be done. I'm wondering how much of existing xvd.rs will be left after we are done with new streaming pipeline

PR #169, opened 2026-09-01
feat: add support for parsing package resource index (Resources.pri)

imLinguin:
Few notes about the implementation
- Use of tokio's `Bytes` across the module
- str slices are wrapped with Arc
- resources specific Cursor type for runtime checking of the array lengths 
- Custom qualifier is not handled
- resources.pri split into multiple files has to be handled by the caller (I have not encountered a game that would use it yet)

From what I gather resources.pri is mainly used for locales in context of PC games. 

Based on research from https://github.com/chausner/PriTools
I used Claude Code to help me generate struct definitons,sample test data based on the research and my findings from PRI files I captured from few games I own.


crates/msixvc-common/src/resources/index/util.rs
```diff
@@ -0,0 +1,224 @@
...
+
+    pub(crate) fn try_read<T>(&mut self, context: &'static str) -> Result<T::Output, PriParseError>
+    where
+        T: BinaryTryParse,
+        PriParseError: From<T::Error>,
+    {
+        T::try_from_slice(self.take_slice(T::SIZE, context)?).map_err(PriParseError::from)
+    }
+
+    pub(crate) fn read_u16(&mut self, context: &'static str) -> Result<u16, PriParseError> {
```
Vidi0: This is not needed, you can use:
```
use msixvc_common::parse::byteorder::little_endian::*;
Cursor::read<U16>()?;
```

crates/msixvc-common/src/resources/index/util.rs
```diff
@@ -0,0 +1,224 @@
...
+        count: usize,
+        context: &'static str,
+    ) -> Result<Vec<u16>, PriParseError> {
+        (0..count).map(|_| self.read_u16(context)).collect()
+    }
+}
+
+/// Decodes a whole byte buffer as UTF-16LE (used for candidate/data-item payloads,
+/// where the slice bounds already delimit the string exactly).
+pub(crate) fn decode_utf16_bytes(bytes: &[u8]) -> String {
```
Vidi0: This isn't needed, since Rust 1.98 there's [from_utf16le_lossy](https://doc.rust-lang.org/std/string/struct.String.html#method.from_utf16le_lossy)

crates/msixvc-common/src/resources/index.rs
Vidi0, line 1: Usually we follow the other module style: this file should be `resources/index.rs`

crates/msixvc-common/src/resources/index/util.rs
```diff
@@ -0,0 +1,224 @@
...
+use crate::parse::{BinaryParse, BinaryTryParse};
+use crate::resources::error::PriParseError;
+
+/// Borrows `len` bytes starting at `offset` out of `data`, or a
+/// [`PriParseError::Truncated`] if the requested range falls outside of `data`.
+pub(crate) fn slice_range<'a>(
+    data: &'a [u8],
+    offset: usize,
+    len: usize,
+    context: &'static str,
```
Vidi0: The context of an error should be added in the next level, where the context is known. Passing a `context` str into a function just for the error message is not optimal, IMO.

Vidi0: What I mean is that the caller of this function should do:
```
slice_range(...).map_err(|e| Error(e, context))
```
This works for every other function in this module which also takes a `context`.

crates/msixvc-common/src/resources/index/util.rs
```diff
@@ -0,0 +1,224 @@
...
+    let start = offset_units.saturating_mul(2).min(block.len());
+    let end = start
+        .saturating_add(count.saturating_mul(2))
+        .min(block.len());
+    decode_utf16_bytes(&block[start..end])
+}
+
+/// Like [`decode_utf16_at`], but for a zero-terminated (`wcharz`) string whose
+/// length isn't given up front - scans for the terminating NUL code unit.
+pub(crate) fn decode_utf16_z(block: &[u8], offset_units: usize) -> String {
```
Vidi0, line 115: Should use `String::from_utf16le_lossy` to simplify the implementation.

crates/msixvc-common/src/resources/structs.rs
```diff
@@ -0,0 +1,1322 @@
...
+            b"mrm_pri1" => VersionId::Pri1,
+            b"mrm_prif" => VersionId::PriF,
+            b"mrm_pri2" => VersionId::Pri2,
+            b"mrm_pri3" => VersionId::Pri3,
+            _ => return Err(PriParseError::UnknownMagic),
+        };
+
+        // uint16 	unknown, zero
+        // uint16 	unknown, one
+        let (_, r) = r.advance::<typenum::U4>();
```
Vidi0: I prefer having this explicit

```suggestion
        let (_unknown0, r) = r.read::<LeU16>(); // zero
        let (_unknown1, r) = r.read::<LeU16>(); // one
```

crates/msixvc-common/src/resources/structs.rs
```diff
@@ -0,0 +1,1322 @@
...
+
+        let (total_file_size, r) = r.read::<LeU32>();
+        let (table_of_contents_offset, r) = r.read::<LeU32>();
+        let (first_section_offset, r) = r.read::<LeU32>();
+        let (number_of_sections, r) = r.read::<LeU16>();
+
+        // uint16 	unknown, 0xFFFF
+        // uint32 	unknown, zero
+        let (_, r) = r.advance::<typenum::U6>();
```
Vidi0: Same as above:

```suggestion
        let (_unknown2, r) = r.read::<LeU16>(); // 0xFFFF
        let (_unknown3, r) = r.read::<LeU32>(); // zero
```

Vidi0 (changes requested):

crates/msixvc-common/src/resources/structs.rs
```diff
@@ -0,0 +1,1322 @@
+use bitflags::bitflags;
+use num_enum::TryFromPrimitive;
+
+use crate::parse::byteorder::little_endian::{U16 as LeU16, U32 as LeU32};
```
Vidi0: In `msixvc::models::xvd::structs`, I imported this like:
```suggestion
use crate::parse::byteorder::little_endian::*;
```

If `typenum` conflicts, you can `use` it and change the name.

Vidi0: If every integer is little endian, it feels more natural to do:

```
r.read::<U32>();
```

crates/msixvc-common/src/resources/structs.rs
```diff
@@ -0,0 +1,1322 @@
...
+    Scale,
+    HomeRegion,
+    TargetSize,
+    LayoutDirection,
+    Theme,
+    AlternateForm,
+    DXFeatureLevel,
+    Configuration,
+    DeviceFamily,
+    Custom,
```
Vidi0: If the order matters, I prefer to be explicit:

```
Language = 0,
Contrast = 1,
Scale = 2,
...
```

crates/msixvc-common/src/resources/structs.rs
```diff
@@ -0,0 +1,1322 @@
...
+}
+
+impl BinaryTryParse for ResourceHeader {
+    type Output = Self;
+    type Size = typenum::U32;
+    type Error = PriParseError;
+
+    fn try_parse<'a>(
+        r: crate::parse::BytesReader<'a, Self::Size>,
+    ) -> Result<(Self::Output, crate::parse::EmptyReader<'a>), Self::Error> {
```
Vidi0:
```suggestion
    ) -> Result<(Self::Output, EmptyReader<'a>), Self::Error> {
```

crates/msixvc-common/src/resources/structs.rs
```diff
@@ -0,0 +1,1322 @@
...
+    pub section_offset: u32,
+    pub section_length: u32,
+}
+
+impl BinaryParse for TableOfContentsEntry {
+    type Output = Self;
+    type Size = typenum::U32;
+    fn parse<'a>(
+        r: crate::parse::BytesReader<'a, Self::Size>,
+    ) -> (Self::Output, crate::parse::EmptyReader<'a>) {
```
Vidi0:
```suggestion
    fn parse<'a>(r: BytesReader<'a, Self::Size>) -> (Self::Output, EmptyReader<'a>) {
```

crates/msixvc-common/src/resources/structs.rs
```diff
@@ -0,0 +1,1322 @@
...
+    pub section_flags: u16,
+    pub section_length: u32,
+}
+
+impl BinaryParse for SectionHeader {
+    type Output = Self;
+    type Size = typenum::U32;
+    fn parse<'a>(
+        r: crate::parse::BytesReader<'a, Self::Size>,
+    ) -> (Self::Output, crate::parse::EmptyReader<'a>) {
```
Vidi0: Same as above, `BytesReader` and `EmptyReader` are already imported.

crates/msixvc-common/src/resources/structs.rs
```diff
@@ -0,0 +1,1322 @@
...
+        let (section_id, r) = r.read::<[u8; 16]>();
+        let (section_qualifier, r) = r.read::<LeU32>();
+        let (flags, r) = r.read::<LeU16>();
+        let (section_flags, r) = r.read::<LeU16>();
+        let (section_length, r) = r.read::<LeU32>();
+        // uint32  unknown, zero
+        let (_, r) = r.advance::<typenum::U4>();
```
Vidi0: When you don't know what the field means but you know that it is an u32, I prefer using:

```
let (_unknown, r) = r.read::<U32>();
```

This way the type is clear, and you don't need a comment.

There's more cases of this below, I'm not going to leave a comment on every one, but I prefer it this way.

crates/msixvc-common/src/resources/structs.rs
```diff
@@ -0,0 +1,1322 @@
...
+#[derive(Debug, Clone, Copy, PartialEq, Eq, TryFromPrimitive)]
+#[repr(u32)]
+pub enum ResourceValueType {
+    String = 0,
+    Path,
+    EmbeddedData,
+    AsciiString,
+    Utf8String,
+    AsciiPath,
+    Utf8Path,
```
Vidi0: Same as above, if the index is needed to convert from a primitive, add it explicitly.

Vidi0 (approved):
Can merge after addressing Clippy's annotations.

PR #170, opened 2026-09-02
fix: Harden Rust version requirement in the crates.

BellezaEmporium:
Right now, only the documentation shows a specific Rust version requirement, but it hasn't been brought up to Cargo's attention that there's a version to lock in.

This small fix aims to lock up the crates and build to Rust 1.98 according to the documentation and prevent the "cannot build X or Y" potential issues we might gather.

Also has a cargo crate update


BellezaEmporium:
Whoopsie daisy, this PR also seem to provide the xal-rs and ntfs dependency bumps I had earlier on. Erh.

BellezaEmporium:
OK, this time it's clean. R2R

imLinguin (approved):
Thank you!

PR #174, opened 2026-09-04
feat: added protobuf example definition for XUserAdd

howdycat:
added in example design for protobuf implementation within services. the idea is that these definitions would also be used by the runtime to send and receive messages between processes, with messages mapping to the appropriate xgameruntime function.

opening as a draft due to some issues compiling due to dependency structure.

### DIFF
--- a/crates/xodus-service/proto/common.proto
+++ b/crates/xodus-service/proto/common.proto
@@ -1,7 +1,7 @@
 /* definitions of container requests and responses and shared enums */
 
 syntax = "proto3";
-package xodus;
+package xodus_proto;
 
 import "xuser.proto";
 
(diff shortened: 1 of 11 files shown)


crates/xodus-service/proto/common.proto
```diff
@@ -0,0 +1,30 @@
...
+  oneof payload {
+    XUserAddResponse xuser_add_res = 10;
+  }
+}
+
+/* hresult, only some response codes */
+/* full codes: https://learn.microsoft.com/en-us/gaming/gdk/docs/reference/live/xbl-hresult-error-codes?view=gdk-2604 */
+enum HRESULT {
+    HRESULT_UNSPECIFIED = 0;
+    HRESULT_S_OK = 1;
```
ChristopherHX, line 29: S_OK is usually 0, not sure if this is a good idea make it 1.

howdycat: Wasn't sure best way around this - HRESULT_UNSPECIFIED has to be 0, but if we want to match the current error convention, we could alias it to S_OK.

howdycat:
will revisit this after service design

PR #177, opened 2026-09-06
doc: architecture diagram and component explanation

howdycat:


imLinguin (approved):
Great for initial components update

PR #182, opened 2026-09-12
msixvc: add `HashTreeStream` 

Vidi0:
As with other of my recent changes, the new `HashTreeStream` is currently not in use. It's a struct that wraps an `AsyncRead`er over the level 0 hash tree and returns a stream of hash entries. The hash level 0 is verified by an in-memory `Box<[HashEntry]>` of the level 1 hash entries. Each hash page is hashed inside the poll function because it's a single hash, so we don't have to worry about blocking the tokio runtime.

The `HashTreeStream` struct is based on two other structs:
 - `PageVerifier`, which stores a list of hashes and is able to verify pages at any index.
 - `PageStream`, which wraps an underlying `AsyncRead`er and returns a stream of pages (`[u8; PAGE_SIZE]`).

I was able to return the hash entries directly from the buffer of `PageStream` by not invalidating the buffer directly after a poll, so this way no allocation other than the page buffer (and the list of level 1 hashes) is needed.

P.S: I finally understand `Pin` and how to write `poll_` methods!


crates/msixvc/src/hash_tree.rs
```diff
@@ -0,0 +1,359 @@
...
+    /// # Panics
+    ///
+    /// If provided `page` and `page_index` is not in-bounds.
+    pub fn verify_page(&self, page: &Page, page_index: usize) -> Result<(), HashMismatch> {
+        assert!(page_index < self.hashes.len());
+
+        let expected_hash = self.hashes[page_index];
+        let hash: HashEntry = *Sha256::digest(page)
+            .first_chunk::<HASH_ENTRY_LENGTH>()
+            .unwrap();
```
ChristopherHX: Please no unwrap if you return result

ChristopherHX (commented):
Not yet understood this.

crates/msixvc/src/hash_tree.rs
```diff
@@ -0,0 +1,359 @@
...
+    page_verifier: PageVerifier,
+
+    remaining_hashes: usize,
+    next_entry_in_page: usize,
+}
+
+impl<R: AsyncRead> HashTreeStream<R> {
+    /// Creates a new [`HashTreeStream`].
+    ///
+    /// # Panics
```
ChristopherHX, line 196: I do not like panics, but my code does as well..

crates/msixvc/src/hash_tree.rs
Vidi0: It's an infallible unwrap, the digest returns 32 bytes but we only care about the first 24. Let me replace it with an expect to make the intent clear.

crates/msixvc/src/hash_tree.rs
Vidi0: Panics are very useful: they indicate a bug in the program. In this case, the caller must ask for the right amount of level 0 hash entries, so if they ask for an incorrect amount there's a bug in the caller code. In these cases I prefer documenting the possible panics in the docs and letting the caller know what guarantees is has to uphold.

I'd replace this with a `Result` if it were part of public API facing code, but this is only intended to be used by other code in the crate, where that invariant will always be upheld (the hash tree parsing logic guarantees it). Returning a Result would only add noise.

Vidi0:
Replaced the unwrap with expect and explained why each error is actually infallible.

crates/msixvc/src/hash_tree.rs
ChristopherHX: > they indicate a bug in the program.

true, but they terminate the service regardless how fatal the error is. e.g. no UI window with error..

Vidi0: IMO, if there's a bug in the program, it's much better to abort to let you know that there's a bug in the code rather than continuing in an inconsistent state.

PR #185, opened 2026-09-14
fix: handle malformed and oversized SOAP responses

### LINKED ISSUE #146: decrypt_soap_encrypted_data panics on responses over 8 KiB and parses the whole buffer instead of the decrypted slice
`decrypt_soap_encrypted_data` decrypts into a fixed 8 KiB stack buffer and `.expect("Failed")`s if the plaintext doesn't fit, so an oversized RST2 response aborts the process instead of returning `RSTError`. It also throws away the decrypted slice and hands the whole buffer to the XML deserializer.

```rust
let (iv, encrypted) = cipher_value.split_at(16);
let iv: &[u8; 16] = iv.try_into().unwrap();
let decryptor = Aes256CbcDec::new(&key.into(), iv.into());
let mut block = [0; 8192];

decryptor
    .decrypt_padded_b2b::<Pkcs7>(encrypted, &mut block)
    .expect("Failed");
let result = std::str::from_utf8(&block).unwrap();
let data = quick_xml::de::from_str::<T>(result)?;
```

All of the below is from unit tests calling the real function (not a reimplementation), with a control test that decrypts and parses a small payload successfully — so the panics are the production path, not a broken harness.

### 1. Hard 8 KiB ceiling → panic (`utils.rs:123`)

Measured threshold: **8191 bytes of plaintext is fine, 8192 panics** with `Failed: Error`. The cutoff is on the *ciphertext*, not the plaintext — `decrypt_padded_b2b` needs the output buffer to hold the input, and PKCS7 appends a full block when the plaintext is already block-aligned.

Why this is worth fixing rather than bumping the constant: the second call site in `rst/request.rs:87` decrypts the **entire SOAP body**, which for a multi-scope request is a `RequestSecurityTokenResponseCollection` holding several tokens. Token size is not small — the code elsewhere assumes a 4096-byte `binary_secret` (`api/live/mod.rs:56`), which is ~5.5 KB once base64'd. So the headroom shrinks as more scopes get requested per call, and there's no check that would turn overflow into an error rather than a panic.

To be explicit about what I have not shown: I have not captured a real response that exceeds the limit — current flows stay under it, which is why this hasn't bitten yet.

### 2. The decrypted length is discarded (`utils.rs:124`)

`decrypt_padded_b2b` returns the plaintext slice; it's dropped, and `from_utf8(&block)` covers all 8192 bytes. The control test shows this directly: a 30-byte plaintext still parses, which means the remaining ~8160 NUL bytes are being handed to `quick_xml`, which happens to tolerate them.

Harmless today, but it means the XML deserializer never sees the actual message boundary, and anything stricter (or any `T` that cares about trailing content) would break on it. Using the returned slice fixes both this and, with a `Vec` sized from the ciphertext, issue 1.

### 3. `split_at(16)` on a short body (`utils.rs:116`)

A `CipherValue` that decodes to fewer than 16 bytes panics with `mid > len` before anything else runs. `split_at_checked` and an `RSTError` variant would cover it.

---

Happy to send a PR — the fix looks like decrypting into a `Vec` sized from the ciphertext and using the returned slice, but the buffer size is your call, so I didn't want to presume. Repro tests are written either way.

kalopsia-space:
## What

Fixes #146 by removing panic paths from SOAP response decryption.

## Changes

- Reject encrypted payloads shorter than the 16-byte IV.
- Allocate the decryption buffer from the ciphertext size instead of a fixed 8 KiB stack buffer.
- Propagate decryption and UTF-8 errors through `RSTError`.
- Parse only the returned plaintext slice.
- Add regression coverage for short payloads and plaintext at the old 8 KiB boundary.

## Verification

- `git diff --check` passes.
- Local Rust tests could not be run because this environment does not have `cargo`, `rustc`, or `rustup` installed.


imLinguin (approved):

kalopsia-space:
@imLinguin Just following up on PR #185. It has been approved since September 16, all checks are passing, and GitHub reports it as mergeable. Could you merge it when convenient? Thanks!

imLinguin:
Thanks!

PR #186, opened 2026-09-16
fix(cli): resume interrupted downloads

### LINKED ISSUE #23: Downloads need resume&caching
Right now if i ask for the same file from the same storeid, xodus happily downloads it again.
Resuming from the moment it was interrupted, including file completion, would make this more robust.

kalopsia-space:
## What

Fixes #23 by making the normal `xodus-cli download` command resume interrupted downloads and skip files that are already complete.

## Changes

- Add resumable downloads using HTTP `Range` requests.
- Skip the request when the destination already matches the expected size.
- Resume from the existing local byte count when a partial file exists.
- Validate `206 Partial Content` and `Content-Range`.
- Safely restart from zero when a server ignores a range request with `200 OK`.
- Preserve partial files after interrupted transfers.
- Verify the final file size before reporting success.
- Return clear download errors instead of panicking.
- Initialize progress from the existing partial-file size.

## Tests

Added local HTTP-server tests covering:

- Fresh download.
- Resuming a partial download.
- Already-complete download without an HTTP request.
- Server without range support.
- Invalid `Content-Range`.
- Final-size mismatch.

## Scope

This PR targets the normal `xodus-cli download` command. The streaming cache path is intentionally unchanged; its broader cache and error-handling work remains separate.

## Verification

- `cargo fmt --all -- --check` passes.
- `cargo test -p xodus-cli commands::download::tests`: 6 passed.
- `cargo test --workspace`: passed.
- `git diff --check`: passes.
- `cargo clippy -p xodus-cli --tests -- -D warnings` reaches existing unrelated warnings/errors in `commands/streaming.rs` and `commands/extract.rs`; no reported errors were from the changed download code.


BellezaEmporium (commented):
A few comments:

- **L46/48** — the condition is computed into `resume` and then repeated verbatim on the next line. `if resume {` should do.

- **Stale resumes** — the size check at L83-88 catches a truncated download, but not a stale one. If the remote file changes between two runs, byte N locally no longer matches byte N remotely; the server still answers 206, you append a suffix from a different version, and the final length still matches. `If-Range: <ETag or Last-Modified>` fixes this — the server returns a full 200 when the resource has changed, and the code already falls back to the truncate path in that case.

- **Partial files** — `size_mismatch_is_reported_after_an_incomplete_response` pins down the fact that a failed download leaves a truncated file under its final name. On the next run that leftover is indistinguishable from a legitimate partial. Better to write to a suffixed name while in flight and rename on completion, the way browsers do (`.part`, `.crdownload`) — `.xodusdownload` would be fine here.

- **Content-Range** — only the `bytes N-` prefix is checked (L55), the total after `/` is ignored. Comparing it against `expected_size` would catch a changed remote file before anything touches the disk.

- **416** — not handled. It shouldn't normally trigger, since L28-34 already bails out when the local file is larger than expected, but if the remote file shrinks the server answers 416 with `Content-Range: bytes */<len>`. Right now that lands in the `!is_success()` branch and just aborts; discarding the local file and restarting would be friendlier.

kalopsia-space:
Implemented the requested hardening in commit `77eeb46`:

- Removed the duplicated resume condition.
- Uses `.xodusdownload` temporary files and renames only after exact-size completion.
- Preserves resumable state after interrupted transfers.
- Persists and sends `If-Range` validators.
- Validates the complete `Content-Range`, including total size.
- Handles HTTP 416 by discarding stale partial state and restarting.

Verification:
- `cargo fmt --all -- --check` passed.
- `cargo test -p xodus-cli commands::download::tests`: 6 passed.
- `cargo test --workspace` passed.
- `git diff --check` passed.

PR #190, opened 2026-09-19
doc: Xodus service design doc

howdycat:
discussion of services


howdycat:
Since we plan to use Tower for services, each service will have a Request and a Response type associated. If the service handles multiple methods that require different inputs or outputs, we can define that service response/request as an enumeration with each key corresponding to a particular service behavior. For example,

```rust

pub enum LoginRequest { 
   Login {...inputs to login method...},
   Logout {...inputs to logout...},
}

pub enum LoginResponse {
   Login {...outputs from login...},
   Logout {...outputs from logout...},
}

impl Service<LoginRequest> for LoginService {
  type Response = LoginResponse
  ....
}
```

Adding a list of the inputs/outputs for each service behavior will help guide this implementation.

imLinguin:
What would the catalog be responsible for? Calls for browsing the store and user's library?  
I imagine the Download calls will be somewhat driven by whatever we decide to do in code in #188

PR #193, opened 2026-09-21
refactor: xodus service with tower

imLinguin:
This starts as a draft, the goal of this is to create a template to work with when implementing new service calls moving forward as well as defining logical services within xodus-service responsible for interaction with subset of features.   
I'll also define a basic lifecycle of the service - background jobs and control of the automatic exit.


howdycat:
Maybe a good start would be determining the inputs/outputs of the service by defining them in protobuf defs? Could document process of adding a service as part of the service design.

imLinguin:
@howdycat Yes that makes sense, I think this can be a topic for #190 and #188 discussions. Here I want to provide a baseline for moving the implementation forward

PR #194, opened 2026-09-21
Add more connected storage notes

ChristopherHX:
Document my findings of implementation.


docs/xbox/titlestorage.md
```diff
@@ -177,3 +177,19 @@ Service Configuration ID
...
 
 MicrosoftGame.config contains the package name, publisher, version, and SCID.
+
+#### x-xbl-lock-ext
+
+Unknown field, not required for blob download.
+
+#### x-xbl-pfn
+
+Required for `GET /savedgames/<containerName>` and works for `GET /` without any
```
ChristopherHX: Probably need to amend this, found an bug here

docs/xbox/titlestorage.md
```diff
@@ -177,3 +181,21 @@ Service Configuration ID
 An ID provided by Microsoft that allows games to access various cloud services.
 
 MicrosoftGame.config contains the package name, publisher, version, and SCID.
+
+SCID is usually not mentioned in MicrosoftGame.config in XGameSave api games, zero extend the hex xbox titleid to a guid for common pc gdk configurations if none is known.
```
vRXDk5U, line 185: Thank you. I didn't know SCID was optional. (https://learn.microsoft.com/en-us/gaming/gdk/docs/reference/system/microsoftgameconfig/elements/microsoftgameconfig-element-scid?view=gdk-2604#requirement)
I still don't think guessing will be necessary though.
The SCID is given by XGameSaveInitializeProvider. (https://learn.microsoft.com/en-us/gaming/gdk/docs/reference/system/xgamesave/functions/xgamesaveinitializeprovider?view=gdk-2604#parameters)
It is stored in the file path. (https://learn.microsoft.com/en-us/gaming/gdk/docs/features/common/game-save/xgamesave?view=gdk-2604#what-file-path-does-xgamesave-save-to)
The SCID can also be obtained from `userstats.xboxlive.com` or `titlehub.xboxlive.com`.
```
curl 'https://titlehub.xboxlive.com/titles/pfn(49827AkuparaGames.RainWorld_mfvawhmf9sssr)/decoration/Scid' \
 -H 'x-xbl-contract-version: 2'
 -H "Authorization: $authorizationHeader" 
 -H 'Accept-Language: en-us'
```
```
{
	"xuid":"<requesting user's xuid>",
	"titles":[
		{
			"titleId":"896928775",
			"pfn":"Microsoft.MinecraftUWP_8wekyb3d8bbwe",
			"bingId":"686834ad-c7a4-47f2-a66e-b0a0fc548c43",
			"serviceConfigId":"4fc10100-5f7a-4470-899b-280835760c07",
			"windowsPhoneProductId":"e90df8bd-dc71-43be-b7ef-648205f09325",
			"name":"Minecraft for Windows",
			"type":"Game",
			"devices":["PC"],
			"displayImage":"http://store-images.s-microsoft.com/image/apps.31326.13510798885735219.3dfb176a-2479-4b17-ab6e-418794f3932d.a8aae002-3be2-4cd4-9ca3-b3e2e7e14882",
			"mediaItemType":"Application",
			"modernTitleId":"896928775",
			"isBundle":false,
			"achievement":null,
			"stats":null,
			"gamePass":null,
			"images":null,
			"titleHistory":null,
			"titleRecord":null,
			"detail":null,
			"friendsWhoPlayed":null,
			"alternateTitleIds":null,
			"contentBoards":null,
			"xboxLiveTier":"Full"
		}
	]
}
```

docs/xbox/titlestorage.md
```diff
@@ -177,3 +181,21 @@ Service Configuration ID
...
+
+#### x-xbl-lock-ext
+
+Unknown field, not required for blob download. Required for upload, is missing create container returns conflict.
+
+#### x-xbl-pfn
+
+Required to match package of scid, otherwise access is rejected.
```
vRXDk5U, line 193: According to Microsoft documentation, saves can be shared across titles but I haven't yet seen a game that does this. (https://learn.microsoft.com/en-us/gaming/gdk/docs/features/common/game-save/game-saves-walkthroughs-and-samples?view=gdk-2604#sharing-game-saves-across-titles)

docs/xbox/titlestorage.md
ChristopherHX: One of my prototype goals was download them without having access to the license. So only
- MicrosoftGame.config
- store listings

are readable most files of msixvc are encrypted.

Yes for XGameSaveInitializeProvider callers we can just use the passed scid. For non standalone data exports...

> serviceConfigId":"4fc10100-5f7a-4470-899b-280835760c07

is that valid for titlestorage? Could be, need to check. Would be nice if we can query it like that.

docs/xbox/titlestorage.md
ChristopherHX: maybe reword this,
```suggestion
Package Family name is required to be assigned to the used scid, otherwise access is rejected.
```

_Actually need to check more, I might only triggered pfn name validation error_

docs/xbox/titlestorage.md
```diff
@@ -59,6 +59,10 @@ GET /
 }
 ```
 
+Field clientFileTime is not part of the json if not set during upload.
```
ChristopherHX, line 62: how to pass continuationToken?

`?continuationToken=token`?

imLinguin: unless a sample is gathered to confirm its hard to guess - the common pattern would be to do a `?after=token` as well

vRXDk5U: `?continuationToken=token` assuming connected storage behaves the same as the public title storage types.
`maxItems` and `skipItems` may also be used but not `continuationToken` and `skipItems` together.
Source: https://github.com/microsoft/xbox-live-api/blob/8486ddd12ed26ec9e6dd755d08688334e09ebfb9/Source/Shared/xsapi_utils.cpp#L90-L115 https://github.com/microsoft/xbox-live-api/blob/8486ddd12ed26ec9e6dd755d08688334e09ebfb9/Source/Services/TitleStorage/title_storage_service.cpp#L666-L677 (Microsoft, MIT)

